PR #97, opened 2011-08-12
C Block support for functions that normally come in begin/end pairs.

erisdev:
This basically amounts to syntax sugar—makes the C++ interface for these sorts of operations more like the Ruby interface. It only works where blocks are available, of course. Observe.

```
graphics().transform(translate(-scrollX, -scrollY), ^{
    allTheThings.draw();
    player.draw();
});

graphics().gl(^{
    glFantastic3DEffects();
    glDoMagic();
    glApplyTrendyBloomFilter();
});
```

I'm not sure this support is compiled in by default on OS X since Gosu is using the 10.4 SDK, but it would be a nice bonus for people who don't mind compiling Gosu themselves. I'm pretty sure Clang supports Blocks on other platforms, too, so I think they are here to stay.

### DIFF
--- a/Gosu/Graphics.hpp
+++ b/Gosu/Graphics.hpp
@@ -29,6 +29,10 @@ namespace Gosu
     Transform scale(double factor);
     Transform scale(double factorX, double factorY, double fromX = 0, double fromY = 0);
     
+#ifdef __BLOCKS__
+    typedef void (^GraphicsBlock)(void);
+#endif
+    
     //! Serves as the target of all drawing and provides primitive drawing
     //! functionality.
     //! Usually created internally by Gosu::Window.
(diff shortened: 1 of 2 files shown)


pmer:
Would C++ lambdas be more viable now in 2017?

jlnr:
Yep, std::function arguments should work just fine for C++ lambdas. Now we can even use std::function without the ::tr1. Now might actually be a good time to add this, as this is a medium-sized C++ interface change, and I want to bump the version to 0.12 when I release MP3 support anyway (due to changed dependencies on Linux).

jlnr:
I've just given it a try and it seems really straightforward, except for the `Graphics::end_gl` function maybe. I'll probably send a PR and close this one soon.

jlnr:
Thanks for the original PR @erisdev and for bumping this @pmer, I've pushed a std::function-based version at #399

PR #106, opened 2011-11-28
Quake2 allocblock

pmer:
Hello,

I am just presenting this here so that a final statement can be made on whether to include this or not.

This branch implements Quake 2's block allocator, which improves Gosu's speed with allocating images.

Cheers,
Jamer

### DIFF
--- a/src/BlockAllocator.cpp
+++ b/src/BlockAllocator.cpp
@@ -1,43 +1,31 @@
 #include "BlockAllocator.hpp"
+#include <algorithm>
 #include <stdexcept>
 #include <vector>
 
 struct Gosu::BlockAllocator::Impl
 {
     unsigned width, height;
+    unsigned avail_w, avail_h;
 
     std::vector<Block> blocks;
-    unsigned first_x, first_y;
-    unsigned max_w, max_h;
-    
-    void mark_block_used(const Block& block, unsigned a_width, unsigned a_height)
+    std::vector<unsigned> used; // Height table.
+
+    // Rebuild the height table where b was allocated.
+    void recalculate_used(Block b)
     {
-        first_x += a_width;
-        if (first_x + a_width >= width) {
-            first_x = 0;
-            first_y += a_height;
-        }
-        blocks.push_back(block);
-    }
-
-    bool is_block_free(const Block& block) const
-    {
-        // right and bottom are exclusive (not part of the block).
-        unsigned right = block.left + block.width;
-        unsigned bottom = block.top + block.height;
-
-        // Block isn't valid.
-        if (right > width || bottom > height) return false;
-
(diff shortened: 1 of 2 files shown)


jlnr:
FWIW here's a public domain library for rect allocation here: https://github.com/nothings/stb/blob/master/stb_rect_pack.h

pmer:
And here's a page that lists a set of algorithms that solve this problem: http://cgi.csc.liv.ac.uk/~epa/surveyhtml.html#toc.2

pmer:
Okay, here's a proposal:

Let W and H be the width and height of whole tiled image in pixels, and let X and Y be the width of an individual tile.

Currently we treat each of the X * Y tiles as an object that needs to be packed. What if, in an effort to reduce the work done, we try to put all the tiles next to each other? That is, since we want a 1-pixel margin around each tile, what if we look for a free space that is (W + X - 1) * (H + Y - 1). Then we place the tiles within this space, since there will be enough room.

This will not pack objects as densely as we do now.

We can take advantage of the fact that these images are related (they're more likely than a normal set of images to either all be used or all not be used) to achieve a second optimization. We can treat the set of X * Y images as a single block in the texture memory when searching for unallocated places in memory when uploading further images, which reduces the computational complexity of future allocations.

In order provide this second optimization, we wouldn't free any tile from texture memory until all related tiles from a single tiled image have been freed.

One negative consequence of this second optimization is that if someone calls `loadTiles()` but only wants to keep a few of the images produced, and they free the rest, we'll continue using more memory than necessary.

pmer:
Not sure if this is useful, but here's an allocater from webrender.

https://github.com/servo/webrender/blob/122ba09/webrender/src/texture_cache.rs#L75-L82

In [the webrender introduction video](https://air.mozilla.org/bay-area-rust-meetup-february-2016/), @pcwalton says it's fast. Not sure this algorithm fits the needs of Gosu games. Should be looked into.

jlnr:
Ah, nice! I like that the comment specifically mentions *deallocation*.

pmer:
I've started working on this, but I'm going pretty slowly. 🙂

jlnr:
Wow, 9 years. Sorry that I've never really reviewed this in depth, but thanks for also opening #534 :+1:

jlnr:
12 years now... and I finally implemented something like this in #658. Thanks again for opening both tickets, the comment in the Rust source code led me in the right direction.

pmer:
Ah, that's wonderful! If we lived in the same town, I'd say let's go out and celebrate. Nicely done! :)

Appreciate the quality of comments in that PR, too.

jlnr:
Thanks :) This was always one of the topics that I really wanted to get back to, but it was certainly a lot more fun to work on this now with proper C++ unit tests and tooling.

PR #193, opened 2013-10-14
More c++11

oli-obk:
where do you want the FontFlags operator implementations? the header file is not quite the right place for it
and what should i do about the other operating systems? i can't test the code there and they definitely won't compile if c++11 is detected (with the changes in this PR, the other one should be fine)

### DIFF
--- a/Gosu/Font.hpp
+++ b/Gosu/Font.hpp
@@ -30,7 +30,7 @@ namespace Gosu
         //! \param fontFlags Flags used to render individual characters of
         //!        the font.
         Font(Graphics& graphics, const std::wstring& fontName,
-            unsigned fontHeight, unsigned fontFlags = ffBold);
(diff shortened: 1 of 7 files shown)


jlnr:
The time has come, I'm finally dropping support for C++03 in #352. Feels relieving, and I am really happy with C++11. C++14 and 17 are nice, but not too relevant for Gosu.

But I just can't figure out what to do with all my enums. `Gosu::AlphaMode::ADD` feels right, but what I'm still not happy with using templates for bitfields (even if they're typesafe). I also don't want my button IDs to be too long: `Gosu::Button::KEYBOARD_LEFT`? `Gosu::Keyboard::LEFT`?

I think I'll stick with plain `enum` and Delphi's abbreviated prefix convention: `Gosu::kbLeft` (or maybe `Gosu::KB_LEFT`) feels like the perfect identifier length to me. Thanks again, but I'm closing this PR for now.

Issue #208, opened 2014-04-20
Draw before Update


PeterCamilleri:
Under windows versions, it seems that the initial draw callback is happening before the initial update callback. This breaks a lot of code that uses update to set positions and other things.

My work around is to add a call to update to the window initialize method, but this still leaves a lot of code not written by me, broken. The mac version does not exhibit this behaviour.

A quick check of the issue base did not reveal this issue as being reported.

PR #210, opened 2014-04-21
Workaround for draw before update on Windows

myokoym:
See #208 on github

This code isn't so good, but it may rescue some code.
What do you think?


jlnr:
Gosu 0.8 has been released and does the right thing :) This PR is obsolete.

PR #211, opened 2014-04-22
Accept options to Zen.init

myokoym:
Continuation of #209. I tried to pass parameters:

``` ruby
init :fullscreen => true do
  ...
end
```

The following is also accepted:

``` ruby
init :width => 640, :height => 480 do
  ...
end
```

How about it?


jlnr:
Thanks for the contribution, but I've decided to drop the Zen mode - #528

PR #262, opened 2015-01-03
Only set gamepad directions for the 1st & 2nd axes

### LINKED ISSUE #261: button_down?(Gosu::GpLeft) and button_down?(Gosu::GpUp) always true w/ Gosu 0.8.x and Xbox 360 controller
I am going through the [Ruby Tutorial](https://github.com/jlnr/gosu/wiki/Ruby-Tutorial), and I tested out the player movement with different controllers, and `button_down?(Gosu::GpLeft)` is returning true, even when the controller left is not being triggered. I am on Mac OS X 10.10.1 using Gosu 0.8.6 with Ruby 2.1.3. I tried this with the PS4 controller plugged in via USB and with the Xbox 360 controller plugged in via USB (with [this driver](https://github.com/d235j/360Controller)).

I am not sure what is causing this (maybe even the code?). When I unplug either controller, the constant left rotation stops.

Here is a gif of the behavior:

constant-left

As you can see, it just continually rotates left.

Here is the `GameWindow#update` that is being used:

``` ruby
class GameWindow < Gosu::Window
  def update
    if button_down?(Gosu::KbLeft) || button_down?(Gosu::GpLeft)
      puts button_down?(Gosu::GpLeft)
      @player.turn_left
    end

    if button_down?(Gosu::KbRight) || button_down?(Gosu::GpRight)
      @player.turn_right
    end

    if button_down?(Gosu::KbUp) || button_down?(Gosu::GpButton0)
      @player.accelerate
    end

    @player.move
  end
end
```

The `@player` instance is just the `Player` class from the tutorial.

Let me know if there is anything I can do to help with this! Gosu has been great so far, and I am really enjoying using it. :smile:

brettchalupa:
I did not build or test this locally yet, but I will give that a shot. I thought it might be useful to try and code what I was thinking the solution might be.

In the scenario where there are more than 2 axes (for examples, left and right triggers), only use the first and second axes for determining the left, right, up, and down gamepad directions.

Closes #261.

### DIFF
--- a/src/Input/Input.cpp
+++ b/src/Input/Input.cpp
@@ -80,17 +80,19 @@ struct Gosu::Input::Impl
             for (int axis = 0; axis < axes; ++axis) {
                 Sint16 value = SDL_JoystickGetAxis(joysticks[i], axis);
 
-                if (value < -(1 << 14)) {
-                    if (axis % 2 == 0)
-                        currentGamepad[gpLeft - gpRangeBegin] = true;
-                    else
-                        currentGamepad[gpUp - gpRangeBegin] = true;
-                }
-                else if (value > +(1 << 14)) {
-                    if (axis % 2 == 0)
-                        currentGamepad[gpRight - gpRangeBegin] = true;
-                    else
-                        currentGamepad[gpDown - gpRangeBegin] = true;
+                if (axis == 0 || axis == 1) {
+                    if (value < -(1 << 14) && ) {
+                        if (axis % 2 == 0)
+                            currentGamepad[gpLeft - gpRangeBegin] = true;
+                        else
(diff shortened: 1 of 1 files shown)


jlnr:
Thanks for the patch, I'll keep the discussion in #261 :)

FYI, you can test this locally using these steps:

```bash
export GOSU_RELEASE_VERSION=0.8.7
rake gem
(sudo) gem install pkg/gosu-0.8.7.gem
```

jlnr:
Fixed by delegating all these decisions to the SDL in https://github.com/jlnr/gosu/commit/452f46e8f8d868c8188d26cff18518a06bf6153b

Still, thanks for the Pull Request and for reporting this issue!

brettchalupa:
:+1: thanks for the fix!

PR #268, opened 2015-02-28
Use existing methods for angle calculations

Zeta21:
Remove unneeded monkey patching of Numeric class, as the method CP::Body#rot can provide the unit angle vector used for applying forces. Use Gosu::gosu_to_radians to set angles.


jlnr:
Hey, sorry for the ridiculous delay, but thanks for the contribution! I've manually applied the substitution in the new `gosu-examples` project:

https://github.com/gosu/gosu-examples/commit/120d0d13aef61e00b51538bf0bd533432d503f67

PR #307, opened 2015-10-26
Make README.txt README.md and update formatting

mhsjlw:
I have renamed README.txt to README.md and update formatting to include links


jlnr:
Thanks! I've looked at the README.md in your account, and I think I'd also turn the irc:// URL into a link. And I think the packaging files need to be updated, as I think they all contain the README.md.
I can quickly go over these myself, but probably not this week :(

mhsjlw:
@jlnr I have tried to make the irc link clickable but, GitHub (from what I know) doesn't allow non http links to be used as hyper links. I could however use the web version of freenode if you'd like.

jlnr:
Oh, thanks, I wasn't aware of this limitation. You are right, irc:// links don't seem to work, not even as direct HTML <a> tags. Web client link sounds good!
Also, all URLs in the README should use https:// instead of http:// (both github.com and libgosu.org), and I think the email link has two extra slashes.
Again, thanks for looking into this long-overdue conversion :)

mhsjlw:
@jlnr No problem, I'll update the PR now

mhsjlw:
Also, yes I made a mistake with the `mailto` link. It is `mailto:` not `mailto://`

mhsjlw:
Done.

jlnr:
Thanks, merged. I'll go ahead with additional edits...parts of the README have actually been outdated :)

PR #324, opened 2016-04-09
Update Compilation.cmake

stantonwjones:
This causes an error. TextMac.cpp does not exist


jlnr:
Why did you close this? Looks like the .cmake file is indeed outdated ^^" Thanks for pointing that out.

stantonwjones:
I closed one of them because the update was misspelled, causing further errors

jlnr:
I'll try to play around with the CMake files as part of #325.

jlnr:
Fixed in #325

PR #339, opened 2016-08-15
Fix Travis CI

jlnr:
Not sure why I still receive errors about -fobjc-arc.


jlnr:
Might have something to do with:
`ARCH_FLAG= -arch x86_64 -arch i386`

PR #347, opened 2016-11-30
first step towards UWP support

Dahrkael:
Almost full UWP support.
- all UWP specific stuff is inside GOSU_IS_UWP macro
- all files that are not Gosu's source code is inside winrt folder
- included SDL2, SDL2_ttf and OpenAL projects with UWP support for an out-of-the-box experience
- x86, x64 and ARM support

Missing: OpenGL ES 2.0 code inside GOSU_IS_OPENGLES, currently 1.1, which is incompatible with SDL2 and ANGLE.


Dahrkael:
i broke unix build, brb fixing

PR #350, opened 2016-12-06
Initial UWP support

Dahrkael:
+ added UWP solution and projects inside winrt folder to keep them contained.
+ added GOSU_IS_UWP macro to detect when compilling for UWP and include/exclude specific code.
+ added specific code for UWP, mainly because incompatibilities between WIN32 and UWP.

OpenAL dependency is another fork with UWP backends, maybe mergeable with the current one.


jlnr:
A belated big Thank You for this PR, but now that SDL is dropping support for UWP, I think it is time to rethink this. The OpenAL dependency has already been replaced in the meantime.

PR #351, opened 2016-12-14
C++ 11 compiler test

jlnr:
Let's see if all compilers in our CI can handle this


jlnr:
We should also kill TR1.hpp and the GOSU_UNIQUE_PTR macros. I would be surprised if this broke anything

Issue #294, opened 2015-08-22
Bring back Image#async_new


jlnr:
Gosu needs more experimental features! - It used to work, and it should be easy to bring it back.

jlnr:
Reminder to look at https://github.com/nothings/stb/issues/309 when this happens

PR #352, opened 2016-12-18
The Great C++11 Refactoring

jlnr:
The C++ codebase really needs a cleanup. Here are some big items I want to achieve:

- [x] Get rid of the directory structure in `src/`. It has caused nothing but problems, especially for `extconf.rb`.
- [x] Remove deprecated parts of the C++ interface.
- [x] Change all method names from camelCase to snake_case to be more consistent with Ruby and the C++ standard library/boost.
- [ ] ~~Reformat everything with clang-format. Curly braces should be unified: always on the same line, except for namespaces/classes/functions.~~ (this will be another PR)
- [x] C++11: Use `nullptr`
- [x] C++11: Use `auto` where appropriate
- [x] C++11: for-each loops
- [ ] ~~C++11: enum classes~~ see #193
- [x] C++11: Remove all references to `tr1` namespace
- [x] Get rid of the `GOSU_IS_CPP11` macro. It should always be treated as defined.
- [x] Use `std::string` instead of `std::wstring`? UTF-8 is a thing now...
- [ ] ~~Kill `unsigned` wherever possible? I hate `unsigned`, but C++ is full of `size_t` anyway and this will cause subtle bugs =/~~ I'm scared, let's don't
- [ ] ~~Drop `beginFoo`/`endFoo` pairs in favour of functions that take functors, as in Ruby.~~ (another PR; similar to the open PR by @erisdev)
- [x] Put `#ifdef GOSU_IS_PLATFORM` guards in every `.cpp` file so that all platforms can just compile `*.cpp`? Maybe better than maintaining lists of source files...
- [ ] ~~Update CONTRIBUTING.md~~ (after clang-format PR)
- [x] Update examples
- [ ] Update Wiki pages


jlnr:
Ouch, I completely broke the Ruby gem (contains only `ruby_wrap.cxx`, no other source files) and the Windows CI didn't notice. Should probably `puts Gosu::Color::NONE` to trigger linker failures, not just `GOSU_VERSION`.

@pmer Since you've contributed the C++ libgosu formula to Homebrew – and do you have any thoughts on this PR? Sorry for breaking backwards compat, it'll be a PITA in my own game too. But if you can think of other changes to the C++ interface, now would be a good time for them :)

pmer:
Wow, this is awesome! Yay for C++11! 😃

You've no worries from me about backward compatibility—I love it. 👍

Hmm, well we have Rvalue refs available now, which might be used in a few places. I think there was an image function I modified years ago back before `std::move` was widely available but which you said would be cleaner if we didn't have to modify... Is that one still around?

And as long as we're making changes, I think it would be nice to fix some of the compiler warnings when including header files/template files. They're kind of noisy, as is... 200 lines on a clean build and some smaller number whenever I change a common header file that a Gosu-using source file includes. https://gist.github.com/anonymous/e37567116ccedfe8ffeb08a938dfd916

Fixing the compiler warnings *would* involve messing with the `unsigned` values, or at least `static_cast<>`ing them. I know what you mean about being scared, though. 😅 Running into similar issues and still trying to work that out for my own game.

And, well, C++ is one of those languages that, like C before it, doesn't have a standard style:

- camelCase: https://github.com/mozilla/gecko-dev/blob/master/gfx/2d/2D.h
- snake_case: https://github.com/BVLC/caffe/blob/master/src/caffe/blob.cpp

So if you're going to use a C++ library, you kind of have to be okay with either style. This is just a long-winded way of me saying snake_case is okay... And the similar names to Ruby makes Gosu feel simpler and more unified! 🙂

Also since the Homebrew formula is brand new I don't think we'll have too many C++ users using it yet, so we shouldn't need to make an additional old-version formula.

pmer:
Also, this one is more a wish that's probably out of scope for what's going on here, but TsunagariC could still a different API for splitting up tiled images. 😉 I still maintain a fork with the (patented? 😓) Quake 2 memory allocation algorithm due to the current O(n^2) algorithm that's in Gosu by default.

The current game I have ends up with about 1000 tiles total, and it might add more as more content is added.

jlnr:
Ah, thanks for reminding me of the warnings, I should really have a look. I could even integrate them into the CI...ouch, that sounds strict.

My thoughts as well about the homebrew formula.

I haven't forgotten #106 😅 The O(n^2) algorithm is just *so* embarrassing. `Gosu::ImageData::subimage` can be used for loading a precompiled texture atlas, if that helps. Another thing that is actually related to C++11: Now that `<thread>` should be available on all platforms (haven't checked yet), Gosu can finally become thread-safe, and we can load large images on a background thread.

jlnr:
Rvalue references: The implicitly generated move-constructors should generally do the right thing (`Bitmap` is being copied around quite a bit). But I'll skim Effective Modern C++ another time before merging this PR just to be sure.

Dahrkael:
how feasible would it be to have all the public classes/methods accept const char* instead of const std::string&? it shouldnt be THAT ugly with little .c_str() and std::string(pointer).
this is for targeting .NET and maybe other platforms that doesnt like templates

pmer:
Loading images with `<thread>` is a really interesting idea, but we might want to be careful about how it's done given that we probably don't want to leak implementation details to games about how their images are loaded. That is, we might not want them to have to be aware that their images are loaded asynchronously—it'd be nice if it were invisible.

We might get a linear increase in speed with the number of cores on the system if we had a multi-threaded texture-packing algorithm.

But if we could reduce the complexity of the O(n²) algorithm that would provide a bigger speedup and might negate the need to run multiple threads. It would likely have a bigger effect and would not be as complicated to implement.

pmer:
@Dahrkael: Not sure about this, but I think there might also be some other templates used in the public interface as well. Not sure if these would be an issue. E.g., `loadTiles()` returns a `std::vector`, etc.

jlnr:
Not to mention that every class has `unique_ptr`/`shared_ptr` template members that a compiler needs to understand to correctly free the object. @Dahrkael Have you tried whether the .NET thingie can really not handle `std::string`? I'd really like to use it, because that way you can pass either `std::string` or `char*` and both will work. And I prefer it for output because it avoids ugly questions about pointer ownership (especially in multi-threaded apps!).

jlnr:
@pmer Oh, I wouldn't make multi-threaded loading automatic. I wouldn't even use `thread` at all, just `mutex` where appropriate. Games can then perform multi-threaded loading themselves using the C++ stdlib (`future` and friends).
In Ruby, I'd probably just make it `async_load_tiles` in addition to `load_tiles`, `Image.async_new` in addition to `Image.new` etc. Parts of that were in the codebase before, but I dropped it because I wasted too much time on `#ifdef`.

jlnr:
Ah right, that's issue #294.

jlnr:
CI is green, TextInput passes some tests, let's merge this.

The wiki updates can wait until this has been released.

PR #358, opened 2017-01-04
[WIP] Resolves #234 - Rework Gosu::VERSION and Gosu::LICENSES

### LINKED ISSUE #234: GOSU_COPYRIGHT_NOTICE should depend on platform
There is a constant called `GOSU_COPYRIGHT_NOTICE` in both Ruby and C++, supposed to be displayed in the game's credits or about screen. Rationale: Depending on the OS, Gosu uses several LGPL/BSD licensed libraries that need to be mentioned to comply with their license. This constant has the following value (as of Gosu 0.8.0):

> This software uses the following third-party libraries:
> 
> Gosu, http://www.libgosu.org, MIT License, http://opensource.org/licenses/MIT
> SDL 2, http://www.libsdl.org, MIT License, http://opensource.org/licenses/MIT
> FreeImage, http://freeimage.sourceforge.net, FreeImage Public License
> libogg & libvorbis, http://www.xiph.org, BSD License, 3-Clause Version, http://www.xiph.org/licenses/bsd
> libsndfile, http://www.mega-nerd.com/libsndfile, GNU LGPL 3, http://www.gnu.org/copyleft/lesser.html
> OpenAL Soft, http://kcat.strangesoft.net/openal.html, GNU LGPL 2, http://www.gnu.org/licenses/old-licenses/lgpl-2.0.html

However, this is not really accurate. Gosu on OS X only uses half of these libraries. Gosu on iOS cannot even use any LGPL-licensed libraries, and printing this GOSU_COPYRIGHT_NOTICE in an iOS game might cause more trouble than it protects against.

TODO: Ensure that `rake/set_version.rb` builds an `#ifdef` cascade that only concatenates the correct strings

Optional: Remove the dependency on `<algorithm>` from `Gosu/Platform.hpp`; `Platform.hpp` and `Version.hpp` are supposed to be super light-weight.

Kjarrigan:
It's a new year and I want to contribute more to the (ruby) community so I choose a (somewhat) random issue I thought wasn't that hard, but in fact I got some (maybe understanding) problems/questions.

I changed the Gosu/Version.hpp to use the constants defined in Gosu/Plattform.hpp to switch the version depending on the platform, but the ruby binding ist generated through swig which interprets the ifdef-else clause based on the platform of the system which calls rake swig. So I actually need to add the ifdef-else clase in the RubyGosu.cxx as well.

Now to the problem - is my understanding of the "problem" correct and if so - how to tell Swig (probably in ruby.i) to preserve the ifdef clause.

The commits in travis and appveyor are atm just for "testing" purposes as I found no real hint how such an change could otherwise be tested in the gosu environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,7 @@ install:
 
 script:
   # Ruby/Gosu
-  - ruby -rgosu -e 'puts Gosu::VERSION'
+  - ruby -rgosu -e 'puts RUBY_VERSION, Gosu::VERSION, Gosu::LICENSES'
+
   # Compile C++ examples
   - mkdir -p examples/build && cd examples/build && cmake .. && make && cd ../..
--- a/Gosu/Version.hpp
+++ b/Gosu/Version.hpp
@@ -3,12 +3,9 @@
 #define GOSU_MAJOR_VERSION 0
 #define GOSU_MINOR_VERSION 10
 #define GOSU_POINT_VERSION 8
-#define GOSU_VERSION "0.10.8"
 
-#define GOSU_COPYRIGHT_NOTICE \
-    "This software uses the following third-party libraries:\n" \
-    "\n" \
-    "Gosu, https://www.libgosu.org, MIT License, http://opensource.org/licenses/MIT\n" \
-    "SDL 2, http://www.libsdl.org, MIT License, http://opensource.org/licenses/MIT\n" \
-    "libsndfile, http://www.mega-nerd.com/libsndfile, GNU LGPL 3, http://www.gnu.org/copyleft/lesser.html\n" \
-    "OpenAL Soft, http://kcat.strangesoft.net/openal.html, GNU LGPL 2, http://www.gnu.org/licenses/old-licenses/lgpl-2.0.html\n"
+namespace Gosu
+{
+    extern const std::string VERSION;
+    extern const std::string LICENSES;
+}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,10 +27,10 @@ after_build:
 
     If ($env:PLATFORM -Eq "x64") {
       $env:PATH = "C:\Ruby23-x64\bin;$OLD_PATH"
-      ruby -I. -r./gosu.rb -e "puts RUBY_VERSION, Gosu::VERSION"
+      ruby -I. -r./gosu.rb -e "puts RUBY_VERSION, Gosu::VERSION, Gosu::LICENSES"
     }
     Else {
       $env:PATH = "C:\Ruby23\bin;$OLD_PATH"
-      ruby -I. -r./gosu.rb -e "puts RUBY_VERSION, Gosu::VERSION"
+      ruby -I. -r./gosu.rb -e "puts RUBY_VERSION, Gosu::VERSION, Gosu::LICENSES"
     }
 test: off
--- a/src/RubyGosu.cxx
+++ b/src/RubyGosu.cxx
@@ -2204,8 +2204,9 @@ namespace Swig {
 #define SWIGTYPE_p_char swig_types[10]
 #define SWIGTYPE_p_double swig_types[11]
 #define SWIGTYPE_p_std__arrayT_double_16_t swig_types[12]
-static swig_type_info *swig_types[14];
-static swig_module_info swig_module = {swig_types, 13, 0, 0, 0, 0};
+#define SWIGTYPE_p_std__string swig_types[13]
+static swig_type_info *swig_types[15];
+static swig_module_info swig_module = {swig_types, 14, 0, 0, 0, 0};
 #define SWIG_TypeQuery(name) SWIG_TypeQueryModule(&swig_module, &swig_module, name)
 #define SWIG_MangledTypeQuery(name) SWIG_MangledTypeQueryModule(&swig_module, &swig_module, name)
 
@@ -2498,43 +2499,6 @@ SWIG_From_int  (int value)
 }
 
 
-SWIGINTERN swig_type_info*
-SWIG_pchar_descriptor(void)
-{
-  static int init = 0;
-  static swig_type_info* info = 0;
-  if (!init) {
-    info = SWIG_TypeQuery("_p_char");
-    init = 1;
-  }
-  return info;
-}
-
-
-SWIGINTERNINLINE VALUE 
-SWIG_FromCharPtrAndSize(const char* carray, size_t size)
-{
-  if (carray) {
-    if (size > LONG_MAX) {
-      swig_type_info* pchar_descriptor = SWIG_pchar_descriptor();
-      return pchar_descriptor ? 
-	SWIG_NewPointerObj(const_cast< char * >(carray), pchar_descriptor, 0) : Qnil;
-    } else {
-      return rb_str_new(carray, static_cast< long >(size));
-    }
-  } else {
-    return Qnil;
-  }
-}
-
-
-SWIGINTERNINLINE VALUE 
-SWIG_FromCharPtr(const char *cptr)
-{ 
-  return SWIG_FromCharPtrAndSize(cptr, (cptr ? strlen(cptr) : 0));
-}
-
-
 SWIGINTERNINLINE VALUE
 SWIG_From_unsigned_SS_long  (unsigned long value)
 {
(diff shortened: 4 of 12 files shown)


.travis.yml
```diff
@@ -57,6 +57,7 @@ before_install:
 branches:
   only:
   - master
+  - gosu-copyright-notice
```
jlnr: Reminder: remove this later when everything works

Gosu/Version.hpp
```diff
@@ -1,14 +1,33 @@
 #pragma once
 
-#define GOSU_MAJOR_VERSION 0
-#define GOSU_MINOR_VERSION 10
-#define GOSU_POINT_VERSION 8
-#define GOSU_VERSION "0.10.8"
+#include <Gosu/Platform.hpp>
 
+#define GOSU_MAJOR_VERSION 9
```
jlnr: Please revert to the current version (0.10.8) for easier merging

jlnr:
Thanks for the contribution! I'd like to propose a few changes that go a little further :) I'm in the process of cleaning up the C++ codebase, and I think that should extend to even the little things like the constants in this patch.

I think the COPYRIGHT_NOTICE should not be a `#define`, it should be a `extern const std::string` that is defined in a new `Version.cpp` file. That solves all the problems with SWIG and `#ifdef`. Its definition does not need to be generated in Ruby either, the whole `#ifdef` logic can just happen inside the `.cpp` file. It's only in `set_version.rb` right now because the script rewrites the whole `Version.hpp` file every time.

I also think it should be called `Gosu::LICENSES` in both Ruby and C++ (-> `namespace`), because it's not actually a [copyright notice](https://en.wikipedia.org/wiki/Copyright_notice). If you update the rdoc (reference/gosu.rb), can you please also fix `Gosu::GOSU_VERSION` -> `Gosu::VERSION`, I've just noticed that.
To preserve backwards compatibility in Ruby, we can just set `Gosu::GOSU_COPYRIGHT_NOTICE = Gosu::LICENSES" in `patches.rb`. Backwards compatibility in C++ is not important, the compiler makes it impossible to miss name changes.

Also, I think GOSU_VERSION should be `const std::string Gosu::VERSION` in C++ too. Only the MINOR/MAJOR/POINT macros are useful for preprocessor magic, anyway. The definition in the `.cpp` file can just `std::to_string` and concat the three macros.

And last but not least, I've learned that MIT-licensed libraries need attribution too according to some interpretations of the license. That means Gosu itself also needs to go into `Gosu::LICENSES`.

Kjarrigan:
Good Morning @jlnr , 

thanks for your feedback. So the unfinished PR was a good idea as I guessed there would be some "other" ideas for an 2 year old ticket ;-) I'll look through your suggestions and report back.

src/RubyGosu.cxx
```diff
@@ -11415,8 +11415,10 @@ SWIGEXPORT void Init_gosu(void) {
   rb_define_const(mGosu, "MAJOR_VERSION", SWIG_From_int(static_cast< int >(0)));
   rb_define_const(mGosu, "MINOR_VERSION", SWIG_From_int(static_cast< int >(10)));
   rb_define_const(mGosu, "POINT_VERSION", SWIG_From_int(static_cast< int >(8)));
-  rb_define_singleton_method(mGosu, "VERSION", VALUEFUNC(_wrap_VERSION_get), 0);
-  rb_define_singleton_method(mGosu, "LICENSES", VALUEFUNC(_wrap_LICENSES_get), 0);
+  // rb_define_singleton_method(mGosu, "VERSION", VALUEFUNC(_wrap_VERSION_get), 0);
+  rb_define_const(mGosu, "VERSION", SWIG_From_std_string(static_cast< std::string >(Gosu::VERSION)));
+  // rb_define_singleton_method(mGosu, "LICENSES", VALUEFUNC(_wrap_LICENSES_get), 0);
+  rb_define_const(mGosu, "LICENSES", SWIG_From_std_string(static_cast< std::string >(Gosu::LICENSES)));
```
Kjarrigan: I'm not sure about this. rake swig generate the singleton_method-wrapper thing which didn't work. If I manually insert what I would expect to be generated it works. But I have no idea what to do...

Kjarrigan:
I'm back, @jlnr and changed some stuff (hopefully) as you requested. There is one issue with swig I didn't know how to handle (see the diff-comment above) but other than that it seems to work.

Gosu/Version.hpp
```diff
@@ -3,12 +3,9 @@
...
-#define GOSU_COPYRIGHT_NOTICE \
-    "This software uses the following third-party libraries:\n" \
-    "\n" \
-    "Gosu, https://www.libgosu.org, MIT License, http://opensource.org/licenses/MIT\n" \
-    "SDL 2, http://www.libsdl.org, MIT License, http://opensource.org/licenses/MIT\n" \
-    "libsndfile, http://www.mega-nerd.com/libsndfile, GNU LGPL 3, http://www.gnu.org/copyleft/lesser.html\n" \
-    "OpenAL Soft, http://kcat.strangesoft.net/openal.html, GNU LGPL 2, http://www.gnu.org/licenses/old-licenses/lgpl-2.0.html\n"
+namespace Gosu
+{
+  extern const std::string VERSION;
```
jlnr: Four spaces :)

appveyor.yml
```diff
@@ -28,9 +28,11 @@ after_build:
     If ($env:PLATFORM -Eq "x64") {
       $env:PATH = "C:\Ruby23-x64\bin;$OLD_PATH"
       ruby -I. -r./gosu.rb -e "puts RUBY_VERSION, Gosu::VERSION"
+      ruby -I. -r./gosu.rb -e "puts Gosu::LICENSES"
```
jlnr: Can probably go on one line - `puts RUBY_VERSION, Gosu::VERSION, Gosu::LICENSES`

src/Version.cpp
```diff
@@ -0,0 +1,32 @@
...
+  const std::string VERSION = std::to_string(GOSU_MAJOR_VERSION) + '.' + std::to_string(GOSU_MAJOR_VERSION) + '.' + std::to_string(GOSU_POINT_VERSION);
+
+#ifdef GOSU_IS_WIN
+  const std::string LICENSES =
+      "This software uses the following third-party libraries:\n"
+      "\n"
+      "Gosu, https://www.libgosu.org, MIT License, http://opensource.org/licenses/MIT\n"
+      "SDL 2, http://www.libsdl.org, MIT License, http://opensource.org/licenses/MIT\n"
+      "libsndfile, http://www.mega-nerd.com/libsndfile, GNU LGPL 3, http://www.gnu.org/copyleft/lesser.html\n"
+      "OpenAL Soft, http://kcat.strangesoft.net/openal.html, GNU LGPL 2, http://www.gnu.org/licenses/old-licenses/lgpl-2.0.html\n";
```
jlnr: It's shorter (more "DRY") to only have one string and `#ifdef` specific lines of it

jlnr:
Ah, I remember that problem with SWIG. Only `#define` counts as a constant, which is dumb: http://www.swig.org/Doc1.3/Ruby.html#Ruby_nn14

I think the workaround is to `%ignore Gosu::LICENSES` and `%constant std::string LICENSES = Gosu::LICENSES`, like in this old commit (not sure if it worked...): https://github.com/gosu/gosu/commit/b7f0c0989396ca28038e217519b79e08697e015e

Can you please give it a try? Otherwise I can look into it, I've gotten good at wrangling SWIG (sadly).

Kjarrigan:
And here I am again :-)
Changed the files as per your request. I hope now finally all check pass. Already rebased against master and ready to squash everything in one commit to finish this issue.

Kjarrigan:
Yeaha. It finally works, although I know what you mean by "wrangling with SWIG". I have never used SWIG before and it just didn't do what I expected. I had to add more to the SWIG_RENAME_PATCH which just feels wrongs, but I got no other ideas atm.

I didn't squashed the travis/appveyor commit as I wasn't sure if it reallys passes and if you want to keep the extra LICENSE output there.

Kjarrigan:
I just found out why I haven't heard of GOSU_COPYRIGHT_NOTICE before I started working at this - because it was in the "TOP LEVEL NAMESPACE" although in Ruby you access it by Gosu::GOSU_COPYRIGHT_NOTICE. 

So I moved it.

jlnr:
No need to worry about squashing, GitHub gives me the choice between merge/squash/rebase when I click the button :) Thanks again for the contribution.

jlnr:
Well, oops. Seconds after posting this, I've clicked the wrong button. Oh well, my bad :)

Issue #304, opened 2015-09-30
Fix the text/string mess


jlnr:
- [ ] ~~Gosu::Font's constructor should take a block that renders glyphs, and []= should be deprecated.~~ delayed until I feel like implementing this
- [x] Make the whole `<b>HTML-like</b>`  formatting more explicit. I think there should be `Font::draw_markup` and `Image.from_markup`, and the existing methods should *not* parse markup (breaking change -> this would mark Gosu 1.0.0!).
- [x] Replace SDL_ttf and other TTF rendering code by `stb_truetype`, so that font metrics are finally the same across platforms.
- [ ] When a TTF file cannot be found, either all platforms should throw an exception, or none.
- [x] When a system font cannot be found, either all platforms should throw an exception, or none.
- [x] Minor issue: The C++ version of `Gosu::createText` should have its parameters shuffled around to be more consistent with the Ruby version.
- [x] Add tests for text rendering and let them run on the CI.
- [ ] ~~Make sure that text entities are as high as the text around them (really? - old idea from my To Do list)~~ Text entities have been removed, use Font[]= instead

jlnr:
Closed via the PRs above.

PR #360, opened 2017-01-06
Test Suite for Gosu Part 1

Kjarrigan:
# ToDo

* [x] Initial Testsetup and Interface test (done by #368)
* [x] Tests for Deprecations (in progress by #370)
* [x] Tests for Gosu::Module
* [x] Tests for Color

### DIFF
--- a/test/test_color.rb
+++ b/test/test_color.rb
@@ -33,18 +33,16 @@ class TestConstants < Minitest::Test
   end
 
   def test_color_attribute_ranges
-    # atm this automatically sets the values to max, min if the exceed the range ...
+    # alpha, red, green, blue are clamped to 0..255
     assert_equal Gosu::Color::WHITE, Gosu::Color.new(300, 300, 300, 300)
     assert_equal Gosu::Color::NONE,  Gosu::Color.new(-50, -50, -50, -50)
 
-    # ... but this does not. I'm actually in favor of raising and exception if either
-    # of the creation methods get invalid values or at least let them behave them# same.
-    assert_equal Gosu::Color::WHITE, Gosu::Color.from_ahsv(500,  2,  2,  2)
-    assert_equal Gosu::Color::NONE,  Gosu::Color.from_ahsv(-50, -5, -5, -5)
+    # hue wraps(!) at 360 so 361 is the same as 1 and not the maximum value 360
+    assert_equal Gosu::Color.from_ahsv(100, 1, 1.0, 1.0), Gosu::Color.from_ahsv(100, 361, 1.0, 1.0)
 
-    # Not sure about this one, but in some programs (including gimp) saturation and value
-    # accept integers in range 0..100 so we might support this?
-    assert_equal Gosu::Color.from_ahsv(255, 300, 100, 100), Gosu::Color.from_ahsv(255, 300, 1.0, 1.0)
+    # saturation and value are clamped to 0.0..1.0 (since #01120e92f7a3)
+    assert_equal Gosu::Color.from_ahsv(100, 1, 1.0, 1.0), Gosu::Color.from_ahsv(100, 361,  2.0,  2.0)
+    assert_equal Gosu::Color.from_ahsv(100, 1, 0.0, 0.0), Gosu::Color.from_ahsv(100, 361, -1.0, -1.0)
   end
 
   def test_dup_and_gl
@@ -52,7 +50,7 @@ class TestConstants < Minitest::Test
     assert_equal 0xff00ff00, Gosu::Color::GREEN.gl
   end
 
-  # introduced by #316, requested by #316 (@shawn42)
+  # introduced by #316, requested by #333 (@shawn42)
   # this is not documented in rdoc/gosu.rb
   def test_alias
     assert Gosu::Color::AQUA.eql?(Gosu::Color::CYAN)
--- a/test/test_color.rb
+++ b/test/test_color.rb
@@ -1,0 +1,61 @@
+require "minitest/autorun"
+require "gosu" unless defined? Gosu
+
+class TestConstants < Minitest::Test
+  def test_predefined_colors
+    [:NONE, :BLACK, :GRAY, :WHITE, :AQUA, :RED, :GREEN, :BLUE, :YELLOW, :FUCHSIA, :CYAN].each do |base_color|
+      assert Gosu::Color.const_defined?(base_color)
+    end
+  end
+
+  def test_color_creation
+    # ARGB
+    assert_equal Gosu::Color::RED,   Gosu::Color.new( 0xff, 0xff, 0x00, 0x00)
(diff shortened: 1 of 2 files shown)


jlnr:
Hah, nice :) I was just thinking about adding tests (finally) for the text rendering stuff (#304). Images can be created and saved without ever opening a window since Gosu 0.9, so there's a variety of stuff that can be tested now (loading tiled images, colour keying, ...).

One idea that I've had was to use yard's parsing API to read the list of all constants from `reference/gosu.rb`, then unload the file (maybe just `remove_const Gosu`? Not sure if that works), load the real Gosu and then compare. Every constant must exist and classes must have the right `public_instance_methods`, etc.

I would have to get rid of a silly hacks in `reference/gosu.rb` though. The button constants are hacked into one line with non-breaking spaces to keep the rdoc pages short, and `attr_accessor :fullscreen?` is not quite correct as the setter is called `fullscreen=`.

I'll look at the other PR first and hopefully release Gosu 0.11.0 tonight.

Kjarrigan:
Just looked through your PR #368 and think its a really good approach. When writing my initial test stump I thought about writing a parser for `reference/gosu.rb` as I totally forgot that its already a (valid) ruby file ;-)

jlnr:
If you want to extend the test for `draw_line` etc. a bit further - these calls *should* work in a `Gosu.record {...}` block. (Or is that still an instance method on the window? I hope not.) There's no way to verify that the recorded image looks correct, though, as you cannot `#save` recorded images.

test/test_gosu_module.rb
```diff
@@ -0,0 +1,112 @@
...
+    assert_match(/^[a-z]{2}/, Gosu.language)
+
+    assert_equal 0, Gosu.fps
+    assert_equal 0, Gosu.milliseconds
+
+    assert_equal 'sans', Gosu.default_font_name
+
+    # TODO: Test if this can be set in the CI, e.g. with xvfb
+    # Link: https://docs.travis-ci.com/user/gui-and-headless-browsers/
+    # otherwise just check for > 0 as done with the available_* values
```
jlnr: `>= available_width`/`_height` would be even better than `> 0`. I don't think these values should be hardcoded, these tests should also work locally.

jlnr:
Feel free to rebase this so that I can merge the existing tests. It doesn't all have to be in one mammoth PR :)

Kjarrigan:
I will but can you look at the color tests? I left some comments here and there where I think the current implementation is wrong (so my test fails) or has some unexpected behaviour.

test/test_color.rb
```diff
@@ -0,0 +1,61 @@
...
+    assert_equal Gosu::Color::WHITE, Gosu::Color.new(300, 300, 300, 300)
+    assert_equal Gosu::Color::NONE,  Gosu::Color.new(-50, -50, -50, -50)
+
+    # ... but this does not. I'm actually in favor of raising and exception if either
+    # of the creation methods get invalid values or at least let them behave them# same.
+    assert_equal Gosu::Color::WHITE, Gosu::Color.from_ahsv(500,  2,  2,  2)
+    assert_equal Gosu::Color::NONE,  Gosu::Color.from_ahsv(-50, -5, -5, -5)
+
+    # Not sure about this one, but in some programs (including gimp) saturation and value
+    # accept integers in range 0..100 so we might support this?
```
jlnr: Overloading on Float vs Integer is too subtle IMHO -> no change for now.

test/test_color.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  def test_color_attribute_ranges
+    # atm this automatically sets the values to max, min if the exceed the range ...
+    assert_equal Gosu::Color::WHITE, Gosu::Color.new(300, 300, 300, 300)
+    assert_equal Gosu::Color::NONE,  Gosu::Color.new(-50, -50, -50, -50)
+
+    # ... but this does not. I'm actually in favor of raising and exception if either
+    # of the creation methods get invalid values or at least let them behave them# same.
```
jlnr: The RGBA setters should not raise exceptions. Harmless mistakes shouldn't crash the game: `color.alpha = 128 + 128 * Math.sin(...)`.
Hue is treated like an angle, and is not clamped but %-ed into the [0; 360) range. I've pushed a short commit to clamp saturation and value. Thanks for pointing this out.

test/test_gosu_module.rb
jlnr: `>= available_width`/`_height` would be even better than `> 0`. I don't think these values should be hardcoded, all tests should also work locally.

test/test_gosu_module.rb
```diff
@@ -0,0 +1,112 @@
...
+  # TODO: Manipulating the current drawing context
+  # clip_to  flush  gl  record  rotate  scale  transform  translate
+
+  def test_misc
+    assert_match(/^[a-z]{2}/, Gosu.language)
+
+    assert_equal 0, Gosu.fps
+    assert_equal 0, Gosu.milliseconds
+
+    assert_equal 'sans', Gosu.default_font_name
```
jlnr: This will fail on Windows/macOS where the default font is "Arial".

Do you know enough about Linux' font loading to say whether "Arial" could be used there as well?

Kjarrigan:
Rebased against master. The only failing test should be #391

test/test_color.rb
```diff
@@ -0,0 +1,59 @@
+require "minitest/autorun"
```
jlnr: Should use new style:
```ruby
# Encoding: UTF-8
require_relative 'test_helper'
```

Kjarrigan:
> I'll fix the NaN, it should return 0.0 because NaNs are such a pain to deal with. 

Yes please.

> I think 0.94 is fine for now, can you please change it to a delta test for now?

"Done". In the failing test I only check the type of the methods not the specific values so I moved hue from Integer to the Float list.

test/test_gosu_module.rb
```diff
@@ -0,0 +1,112 @@
...
+class TestGosuModule < Minitest::Test
+  def test_drawing_primitives
+    {
+      # args: [x, y, width, height, c, z, mode]
+      draw_rect: [0, 0, 10, 10, 0xff_ffffff, 0, :default],
+
+      # args: [x, y, c]{2,4}, z, mode
+      draw_line: [0, 0, 0xff_ffffff, 10, 10, 0xff_ffffff, 0, :default],
+      draw_triangle: [0, 0, 0xff_ffffff, 10, 10, 0xff_ffffff, 0, 10, 0xff_ffffff, 0, :default],
+      draw_quad: [0, 0, 0xff_ffffff, 10, 10, 0xff_ffffff, 0, 10, 0xff_ffffff, 0, 10, 0xff_ffffff0, 0, :default],
```
jlnr: Test is failing because of the extra 0 behind `0xff_ffffff`

Kjarrigan: Thats definitely a typo :cry:

PR #370, opened 2017-01-09
Add deprecations to Ruby/Gosu (Resolves #367)

### LINKED ISSUE #367: Ruby/Gosu needs deprecations
Ideally, Gosu's C++ core should only support the canonical way to do everything, and `patches.rb` should be renamed to `compat.rb` and provide all the backwards compatibility code. The first time a deprecated feature is used, a warning should be printed.

The goal is not to get rid of the compat code (it's trivial to keep it in Ruby), but to ensure that existing Gosu projects converge towards a coherent, modern style of using Gosu.

Kjarrigan:
I splitted some parts of `patches.rb` into `compat.rb`. Now patches has all the real "Patch" stuff and compat everything related to deprecations. I also added a test although the test branch isn't merged yet so I temporarily enabled a run of this test file in appveyor and travis in the second commit.

I wasn't sure if the Button constants are meant to be deprecated or just as alternative. At the moment I treat them as deprecated, because they are ruby-like enough :smile: 

EDIT: Why is there no CI run...

### DIFF
--- a/lib/gosu/compat.rb
+++ b/lib/gosu/compat.rb
@@ -1,0 +1,179 @@
+# This whole file is about backward compatibility
+
+# First - define some methods to tell the world about their old code
+
+module Gosu
+  DEPRECATION_STACKTRACE_LINES = 1
+
+  # Basically copy&paste from Gem but remove the date-part.
+  def self.deprecate(klass, name, repl)
+    klass.class_eval {
+      old = "_deprecated_#{name}"
+      alias_method old, name
+      define_method name do |*args, &block|
+        Gosu.deprecation_message(self, name, repl)
+        send old, *args, &block
+      end
+    }
+  end
+
+  def self.deprecate_const(name, repl)
+    send(:remove_const, name) if const_defined?(name)
+
+    @@_deprecated_constants ||= {}
+    @@_deprecated_constants[name] = repl
+  end
+
+  # The advantage of this approach is that the old Names won't show up in
+  # any Codecompletion (like IRB) or *.constants, so you have to know the
+  # exact old name to acutally use the old const.
+  # However define the value after the first call, to show the warning only once.
+  def self.const_missing(const_name)
+    if @@_deprecated_constants && repl = @@_deprecated_constants[const_name]
+      Gosu.deprecation_message(self, const_name, repl)
+      const_get(repl)
+    else
+      super
+    end
+  end
+
+  def self.deprecation_message(klass_or_full_message, name=nil, repl=nil)
+    @@_deprecations_shown ||= {}
+
+    msg = if klass_or_full_message.is_a?(String) && name.nil? && repl.nil?
+      [ "DEPRECATION WARNING: #{klass_or_full_message},
+        \n#Called from #{caller[1, DEPRECATION_STACKTRACE_LINES]}"
+      ]
+    else
+      # class-method deprecation warnings result in something like this: #<Class:Gosu::Window>::button_id_to_char is deprecated
+      # so remove the instance-inspect stuff to make it look a bit better
+      target = klass_or_full_message.kind_of?(Module) ? "#{klass_or_full_message.to_s.gsub(/#<Class:(.*)>/, '\1')}::" : "#{klass_or_full_message.class}#"
+      [ "DEPRECATION WARNING: #{target}#{name} is deprecated",
+        repl == :none ? " with no replacement." : "; use #{repl} instead.",
+        "\n#{target}#{name} called from #{Gosu.deprecation_caller.join("\n")}",
+      ]
+    end
+    return if @@_deprecations_shown.has_key?(msg[0])
+    @@_deprecations_shown[msg[0]] = true
+
+    warn "#{msg.join}."
+  end
+
+  # this method removes the noise of the caller due to the deprecation methods
+  def self.deprecation_caller
+    caller.delete_if { |trace_line| trace_line =~ /(deprecat|const_missing)/ }
+      .first(DEPRECATION_STACKTRACE_LINES)
+  end
+end
+
+# No need to pass a Window to Image, Sample and Song
+class Gosu::Image
+  alias initialize_without_window initialize
+
+  def initialize(*args)
+    if args[0].is_a? Gosu::Window
+      Gosu.deprecation_message("Passing a Window to Image#initialize has been deprecated in Gosu 0.9 and the arguments have changed a bit as well. Check out https://www.libgosu.org/rdoc/Gosu/Image.html for more information.")
+      if args.size == 7
+        initialize_without_window args[1], :tileable => args[2], :rect => args[3..-1]
+      else
+        initialize_without_window args[1], :tileable => args[2]
+      end
+    else
+      initialize_without_window(*args)
+    end
+  end
+
+  class << self
+    alias from_text_without_window from_text
+  end
+
+  def self.from_text(*args)
+    if args.size == 4
+      Gosu.deprecation_message("Passing a Window to Image.from_text has been deprecated in Gosu 0.9 and the arguments have changed a bit as well. Check out https://www.libgosu.org/rdoc/Gosu/Image.html for more information.")
+      from_text_without_window(args[1], args[3], :font => args[2])
+    elsif args.size == 7
+      Gosu.deprecation_message("Passing a Window to from_text has been deprecated in Gosu 0.9.")
+      from_text_without_window(args[1], args[3], :font => args[2],
+        :spacing => args[4], :width => args[5], :align => args[6])
+    else
+      from_text_without_window(*args)
+    end
+  end
+end
+
+class Gosu::Sample
+  alias initialize_without_window initialize
+
+  def initialize(*args)
+    if args.first.is_a? Gosu::Window
+      args.shift
+      Gosu.deprecation_message("Passing a Window to Sample#initialize has been deprecated in Gosu 0.7.17.")
+    end
+    initialize_without_window(*args)
+  end
+end
+
+class Gosu::Song
+  alias initialize_without_window initialize
+
+  def initialize(*args)
+    if args.first.is_a? Gosu::Window
+      args.shift
+      Gosu.deprecation_message("Passing a Window to Song#initialize has been deprecated in Gosu 0.7.17.")
+    end
+    initialize_without_window(*args)
+  end
+end
+
+# Moved some Window-methods to the Gosu::Module
+class Gosu::Window
+  
+  # Class methods that have been turned into module methods.
+  class << self
+    def button_id_to_char(id)
+      Gosu.button_id_to_char(id)
+    end
+    Gosu.deprecate self, :button_id_to_char, 'Gosu.button_id_to_char'
+
+    def char_to_button_id(ch)
+      Gosu.char_to_button_id(ch)
+    end
+    Gosu.deprecate self, :char_to_button_id, 'Gosu.char_to_button_id'
+  end
+
+  # Instance methods that have been turned into module methods.
+  %w(draw_line draw_triangle draw_quad
+     flush gl clip_to record
+     transform translate rotate scale
+     button_id_to_char char_to_button_id button_down?).each do |method|
+    define_method method.to_sym do |*args, &block|
+      Gosu.send method, *args, &block
+    end
+    Gosu.deprecate self, method, "Gosu.#{method}"
+  end
+end
+
+# Constants
+module Gosu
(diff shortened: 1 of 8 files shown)


jlnr:
Looks great, thanks! I'll be busy for a few more days, will hopefully catch up with the PRs starting on Friday :)

Kjarrigan:
No hurry. The PR's won't run away :smiley: 

I'll rebase after you merged the initial unit-test, then my CI runs should be green as well.

Kjarrigan:
I rebased and changed a bit here and there to fit in the brand new test-suite. But the biggest change ist that I swapped new and old constants. I thought that Kb0 is the new version and KB_0 the old (because thats the name shown at libgosu.org/rdoc/), but your new test-suite comments made it clear that its the other way around :smile:
Now lets wait for the CI results...

test/test_deprecations.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    assert_output("", /DEPRECATION WARNING: Passing a Window to Image.from_text has been deprecated in Gosu 0.9./) do
+      assert_raises(::TypeError) { Gosu::Image.from_text(win,2,3,4) }
+      assert_raises(::TypeError) { Gosu::Image.from_text(win,2,3,4,5,6,7) }
+    end
+  end
+
+  def test_moved_window_to_gosu_module_methods
+    # Class methods that have been turned into module methods.
+    assert_output("", /DEPRECATION WARNING: Gosu::Window::button_id_to_char is deprecated; use Gosu.button_id_to_char instead./) { Gosu::Window.button_id_to_char(Gosu::KbA) }
```
jlnr: `KbA` -> `KB_A`

jlnr:
I've invited you to the Gosu project for future PRs - it's easier to work with branches when they're all in the same repo 😃

I still have to run this branch, but I think it prints a message every time a deprecated constant or method is used. I think it'd be cool if every constant would only result in one message, and maybe with a stacktrace? That'd make it easier for users to upgrade their codebase. Maybe override `const_missing`, set the constant there but print a message? That way, each message would appear only once because the const would not be missing afterwards.

jlnr:
Re: Failed tests on Travis, maybe silently swallow all exceptions that match ´Could not initialize SDL Video`? I think running these tests on macOS and Windows is good enough :)

Kjarrigan:
Yeaha. All green now. Theresa was already a stacktrace but I reworked that a bit. Can't test it though because the minitest call is so deeply nested that the origin isnt in the caller array...
What do you think?

lib/gosu/compat.rb
```diff
@@ -0,0 +1,187 @@
...
+    end
+  end
+end
+
+# No need to pass a Window to Image, Sample and Song
+class Gosu::Image
+  alias initialize_without_window initialize
+
+  def initialize(*args)
+    if args[0].is_a? Gosu::Window then
```
jlnr: I've given up on my beloved `then` keyword recently. Feel free to delete all occurences of it.

lib/gosu/compat.rb
```diff
@@ -0,0 +1,187 @@
...
+    def self.message(msg)
+      msg = [ "DEPRECATION WARNING: #{msg}",
+        "\n#Called from #{caller[1, WARNING_STACKTRACE_LINES]}",
+      ]
+      warn "#{msg.join}."
+    end
+
+    # this method removes the noise of the caller due to the deprecation methods
+    def self.filtered_caller
+      caller.delete_if { |trace_line| trace_line =~ /(deprecat|const_missing)/ }[0, WARNING_STACKTRACE_LINES]
```
jlnr: `}[` on the same line is a bit too much IMHO :) Maybe `.first(WARNING_STACKTRACE_LINES)`?

lib/gosu/compat.rb
```diff
@@ -0,0 +1,187 @@
+# This whole file is about backward compatibility
+
+# First - define some methods to tell the world about their old code
+module Gosu
+  module Deprecate
```
jlnr: Hmm, `Gem::Deprecate` works the same, but do you know if there is an advantage in having a separate module for deprecations? Can't we just define all the methods on `Gosu`? Then there'd be one less exception in `test_interface`.

lib/gosu/compat.rb
```diff
@@ -0,0 +1,187 @@
...
+      klass.class_eval {
+        old = "_deprecated_#{name}"
+        alias_method old, name
+        define_method name do |*args, &block|
+          Gosu::Deprecate.deprecation_message(self, name, repl)
+          send old, *args, &block
+        end
+      }
+    end
+    def deprecate name, repl
```
jlnr: +`()` for consistency with the other methods

lib/gosu/compat.rb
```diff
@@ -0,0 +1,187 @@
...
+  end
+end
+
+# No need to pass a Window to Image, Sample and Song
+class Gosu::Image
+  alias initialize_without_window initialize
+
+  def initialize(*args)
+    if args[0].is_a? Gosu::Window then
+      Gosu::Deprecate.message("Passing a Window to Image#initialize has been deprecated in Gosu 0.9.")
```
jlnr: This is a little misleading because not only the `window` argument has changed, but some arguments have moved into a hash as well. Not sure what to do about this - do you think a link to the rdoc would be too much?

jlnr:
I left a few nitpicks, I hope I can actually try it later tonight or tomorrow. I guess I should hurry up with Gosu 0.11.1 after 0.11.0 broke so much :) (4-bit PNG files and Unicode on Windows; oops)

lib/gosu/compat.rb
Kjarrigan: I actually moved the code a few time forth and back between a seperate modul and Gosu itself - not sure which version I like more. The final decision was based on the fact, that I wanted the const_missing hook easily available in every class (if needed) and I felt quite uncomfortable extending the Gosu-module, as there is quite some stuff defined. As there are currently only (deprecated) constants und Gosu it would work I think.

lib/gosu/compat.rb
Kjarrigan: Definitely unusual - can't remember that I've seen this lately but that doesn't mean we couldn't do that. We want to help the gosu-folks to switch to the newer version so why not help them by leading them in the right direction (and providing them a rdoc-link)

jlnr:
OK, I've finally checked it out and ran `gosu-examples` with it. Works great, but it's noisier than I thought. Maybe the deprecation message functions should remember which stacktraces have already been printed and only print each location once?

I've also realised that deprecating `Window#translate` (and similar methods) is a bit confusing because when you `include Gosu`, you might end up calling `Window#translate` instead of `Gosu.translate` if you just call `translate(...)` in `MyWindowSubclass#draw`. Will have to sleep over that. Maybe we can just exclude the methods that have moved from `class Window` to `module Gosu` for this version.

jlnr:
Related discussion: https://www.libgosu.org/cgi-bin/mwf/topic_show.pl?pid=9088#pid9088

Kjarrigan:
Hi @jlnr ,
I'm back @ work from my short parental holiday and I now have some routine handling two little girls, so I think I'll soon restart working at my open PR's. I'll look through the forum, reddit and git-issues to get up-to-date and then go on. Anything I need to know ? :smile:

jlnr:
:+1: Grats! I probably won't be very active until around 2-27 though, unless there are any rainy days :)

I've already used this branch to sanitise the `gosu-examples` gem, which runs without deprecation warnings now. But the deprecation of `Window#translate` etc. do indeed cause warnings when you `include Gosu` at the top of the script, as I've done in a little tool last month.

Kjarrigan:
Just rebased and fixed some of the "old" stuff. Next step is to show each deprecation only once and think about the whole include-confusion thingy.

jlnr:
I wonder if an approach based on method_missing would fix the include ambiguity issue. If missing, then issue warning and define_method. Just a quick idea, I'm not at my laptop right now ^^"

Sent from my iPhone

Kjarrigan:
That's actually a great idea. So basically the same approach as with the constants....I'll definitely try that.

Kjarrigan:
I thought about the method missing approach and tried some things, but its way more complicated and more LoC than the current approach so I just added a simple message already show mechanism.

But on my local machine I get random failing tests, which is really odd, because there are no conditions. Either it is deprecated or not. No real magic there. So I think its related to the randomness of the minitest environment or my local setup. Do you have experienced something similiar?

For example. Every other _rake test_ run it shows no deprecation message for constants:

```
1) Failure:

TestDeprecations#test_gosu_module_constants [./gosu/test/test_deprecations.rb:13]: 
In stderr.
Expected /DEPRECATION WARNING: Gosu::Kb0 is deprecated; use KB_0 instead./ to match "".
```

Kjarrigan:
I figured it out...its minitest as assumed. The testsuite nests it's calls so weird that I usually don't get the real caller even with max stack trace lines and so it often occurs that the message even for different tests is recognised as the same origin. So now every message only gets shown once. No matter where it was called.
I guess thats enough, as I personally would seek with _git grep XYZ_ for every occurrence of the relevant text.

test/test_deprecations.rb
```diff
@@ -0,0 +1,75 @@
...
+          assert_raises(::ArgumentError) { win.send(meth) }
+        end
+      end
+    end
+  end
+
+  # Catch a linux specific error
+  def with_gosu_window
+    yield Gosu::Window.new(10,10)
+  rescue RuntimeError => e
```
Kjarrigan: Can be removed if #383 gets merged.

jlnr: Done :)

lib/gosu/compat.rb
```diff
@@ -0,0 +1,179 @@
...
+    @@_deprecations_shown ||= {}
+
+    msg = if klass_or_full_message.is_a?(String) && name.nil? && repl.nil?
+      [ "DEPRECATION WARNING: #{klass_or_full_message},
+        \n#Called from #{caller[1, DEPRECATION_STACKTRACE_LINES]}"
+      ]
+    else
+      # class-method deprecation warnings result in something like this: #<Class:Gosu::Window>::button_id_to_char is deprecated
+      # so remove the instance-inspect stuff to make it look a bit better
+      target = klass_or_full_message.kind_of?(Module) ? "#{klass_or_full_message.to_s.gsub(/#<Class:(.*)>/, '\1')}::" : "#{klass_or_full_message.class}#"
```
jlnr: Please use a multi-line `if` here, this line is really long.

lib/gosu/compat.rb
```diff
@@ -0,0 +1,179 @@
...
+
+  def self.deprecation_message(klass_or_full_message, name=nil, repl=nil)
+    @@_deprecations_shown ||= {}
+
+    msg = if klass_or_full_message.is_a?(String) && name.nil? && repl.nil?
+      [ "DEPRECATION WARNING: #{klass_or_full_message},
+        \n#Called from #{caller[1, DEPRECATION_STACKTRACE_LINES]}"
+      ]
+    else
+      # class-method deprecation warnings result in something like this: #<Class:Gosu::Window>::button_id_to_char is deprecated
```
jlnr: Please add a linebreak here to keep the line a little shorter.

lib/gosu/compat.rb
```diff
@@ -0,0 +1,179 @@
...
+      const_get(repl)
+    else
+      super
+    end
+  end
+
+  def self.deprecation_message(klass_or_full_message, name=nil, repl=nil)
+    @@_deprecations_shown ||= {}
+
+    msg = if klass_or_full_message.is_a?(String) && name.nil? && repl.nil?
```
jlnr: Gosu's own codebase uses `and` instead of `&&` in conditionals (same in `const_missing`). Maybe I'll change my mind in the future, but for now let's keep it consistent.

lib/gosu/compat.rb
```diff
@@ -0,0 +1,179 @@
+# This whole file is about backward compatibility
+
+# First - define some methods to tell the world about their old code
```
jlnr: "...to deprecate methods and classes" might be a little clearer :)

jlnr:
Most of this has been merged in #389. I'm closing this PR because the comment history is pretty long and I still don't see a good way to deprecate the remaining methods without breaking compatibility in edge cases. If you have an idea let me know, otherwise we'll have to wait until 1.0 at least :)

Kjarrigan:
Hi @jlnr, no problem. As long as we keep solving some of the existing "problems" and my spare time at the moment is pretty limited, but I hope to return soon to contribute more.

Issue #369, opened 2017-01-09
When going fullscreen->windowed, the window is not centred


jlnr:
Reported here: https://www.reddit.com/r/gosu/comments/5meny4/gosu_0110_released/dc5jvi1/

Issue #361, opened 2017-01-07
Alt+Enter / cmd+[ctrl+]F should toggle fullscreen by default


jlnr:
macOS: cmd+F is the default shortcut to toggle fullscreen in video players, cmd+ctrl+F is a common shortcut to enable fullscreen in other desktop apps (Safari, Mail, Finder, …).

Windows (and maybe Linux?): Alt+Enter is the default shortcut to toggle fullscreen in games.

Gosu games should support this by default because not every user is aware of each platform's conventions. I'm not sure if Gosu should always enforce these shortcuts, or if it should be the default implementation of `Window#button_down` which people can decide to invoke via `super` or not.

Kjarrigan:
Hah, yesterday I thought maybe it would be a good idea to have a default binding for the gosu licenses (like F1). So I like your proposal of some default key bindings, but prefer the ```super``` version.

jlnr:
Good points. I agree that it needs to be documented. I have added rdoc for the new default implementation, and I will update the tutorial wiki page to explain the default shortcuts too.

jlnr:
Hmmm, I'm not sure where to show the LICENSES by default. The idea is that you should include them in your credits screen (assuming there is one...), which is what I do in my iOS game. I'm not sure whether hiding the attributions behind a F1 default shortcut is sufficient if you don't also document that shortcut in your game. And then you might as well just print the licenses in your game.

Kjarrigan:
But we should definitely to make it clearer that you should include them somewhere, because as I said. Until I worked on that PR I didn't know about that. Maybe instead of an default binding extends the rdoc and add it to the Captain Ruby Example? Who ever starts with gosu will likely use that as first reference. But it might be worth a seperate issue then.

jlnr:
@Kjarrigan Maybe it should just have its own sidebar entry in the `gosu-examples` gem to show the LICENSES string and explain its use at the same time. I'll have to update the `gosu-examples` gem anyway, it uses lots of deprecated stuff.

Hi2885:
Add code

PR #372, opened 2017-01-13
Fullscreen shortcuts (#361) and window placement (#369)

jlnr:


jlnr:
Works great on macOS and Windows 7.

In Fedora 25, fullscreen mode is kinda broken with Xorg. The window is always placed in the top-left corner both with Xorg and Wayland. Not sure if these bugs are caused by running Linux in a (hardware-accelerated) VM? Merging this because I guess these are not new issues.

PR #381, opened 2017-03-08
Resolves #287 - add a SDL_QUIT callback ...

### LINKED ISSUE #287: Callback when closing the window via [x]
Since all platforms use SDL 2 now, this should be either trivial to implement - or impossible - but nothing in-between. ;)

First idea for the interface:

``` ruby
class Gosu::Window
  ...
  def try_close
    close # feel free to override this and do something else instead
  end
  ...
end
```

I'm not really happy with `try_close` though =/

Kjarrigan:
... but not with the close! approach as I got a headache thinking about how to make it work. 

So this is only a small change and after struggling with the real code I'm in favor of using the close? (or shall_close?) keyword. For me its similiar to needs_redraw? which context dependent says yes or no.

This breaks no API but triggers __close?__ too, if Window::close is called.

Trying to make close! I got "a ton" of changes which would break the API for the C++ Version and which would have been rather ugly, but I that definitely could be due to my lack of knowledge around the Gosu-Internals / SWIG and C++  which hopefully reduces over time while trying to resolve some open issues. So far I have learned a bit more how certain things work :smile:

The main problems with the close! approach I got in my head were the following: 

At the moment you call close from Ruby which calls close in C++ which registers a QUIT event which eventually will be processed by Window::tick. Now we want to change this behavior, so that Window::tick calls close() when it receives an QUIT event. So wee need a new method in C++ to register an event (e.g. quit). And close has to return a bool so that I can quit the Window::tick loop. And where should I attach close! ? Should it trigger the quit method which would againg trigger close or should it directly close the window by calling SDL_HideWindow but then how to stop the tick loop from executing? Or do we need a new Event like GOSU_QUIT which is set by close which then is processed by the loop.... :confused:

### DIFF
--- a/Gosu/Window.hpp
+++ b/Gosu/Window.hpp
@@ -77,10 +77,9 @@ namespace Gosu
         virtual bool tick();
         
         //! Closes the window if it is currently shown.
-        void close();
-
-        //! Gives the game a chance to say no to being closed. True by default.
-        virtual bool shall_close() const { return true; }
+        //! If you do not want the window to close immediately, you should override this method and
+        //! only call the base implementation (Window::close) when needed.
+        virtual void close();
 
         //! Called every update_interval milliseconds while the window is being shown.
         //! Your application's main game logic goes here.
--- a/Gosu/Window.hpp
+++ b/Gosu/Window.hpp
@@ -79,6 +79,9 @@ namespace Gosu
         //! Closes the window if it is currently shown.
         void close();
 
+        //! Gives the game a chance to say no to being closed. True by default.
+        virtual bool shall_close() const { return true; }
+
         //! Called every update_interval milliseconds while the window is being shown.
         //! Your application's main game logic goes here.
         virtual void update() {}
--- a/examples/Tutorial/Tutorial.cpp
+++ b/examples/Tutorial/Tutorial.cpp
@@ -193,7 +193,7 @@ public:
     void button_down(Gosu::Button button) override
     {
         if (button == Gosu::KB_ESCAPE) {
-           close();
+            close();
         }
         else {
             Window::button_down(button);
--- a/ext/gosu/gosu.i
+++ b/ext/gosu/gosu.i
@@ -919,7 +919,7 @@ namespace Gosu
 %rename("mouse_y=") set_mouse_y;
 %rename("needs_cursor?") needs_cursor;
 %rename("needs_redraw?") needs_redraw;
-%rename("close?") shall_close;
+%rename("close!") force_close;
 %rename("fullscreen?") fullscreen;
 %markfunc Gosu::Window "mark_window";
 %include "../../Gosu/Window.hpp"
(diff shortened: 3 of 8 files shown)


Kjarrigan:
Will work if #383  gets merged :smile:

jlnr:
I've rebased this branch and pushed a new commit. I tried to stick with `close`/`close!` instead of `close?`/`close`, which I think is a bit simpler, but I also ran into a problem with the way `show` and `tick` interact. I've changed the test to illustrate the remaining problem. It all works if you use `show`, but not if you call `tick` and `close!` on a window instance from the outside.

I wish I knew if anyone actually uses `tick` to integrate Gosu with a foreign main loop (probably a GUI library). Oh well, maybe it's only a matter of adding another boolean flag in Window.cpp?

jlnr:
Because I couldn't come up with a good name for `close!` in C++, `virtual void Window::close` is both the callback *and* the method that closes the window. C++ makes it pretty easy to call specific versions of an overridden method, so you can do this:

```cpp
class W : Window {
...
    void close() override { /* not calling base impl */ }
    void button_down(Button btn) { if (btn == KB_ESCAPE) Window::close(); /* explicitly call base impl; same as close! in Ruby */ }
}
```

jlnr:
One related thing that's on my list is that it would be pretty cool to not call `draw` and the other window callbacks after the user calls `close!`, just to reduce surprises. That could be achieved by polling the window's visibility before every single callback that's run from `Window::tick`, but that's definitely not pretty. Do you think it is good/intuitive enough that callbacks can still be called when a window is dismissed?

jlnr:
Feel free to undo parts of my commit, I've been meaning to finish this today but then I ran out of tinkering time again :(

Kjarrigan:
> I didn't leave the original test in because calling stuff from another thread is not really safe in Gosu. But I just realised that it is possible to test show by calling close/close! from inside def update rather than from the test code (for example, after 5 ticks)

That's what I tried first - calling it in draw/update, but it didn't work as i expected. But I could try again. New weeks so maybe new ideas :smile:

> Because I couldn't come up with a good name for __close!__

Me neither - I used __quit__ / __shutdown__ / __stop__ in some of my test-runs. Too sad C++ doesn't support ! / ?

> Do you think it is good/intuitive enough that callbacks can still be called when a window is dismissed? Nobody has ever complained about it, but it seems odd to me.

I have no problem with that behavior - I express my wish to close the game window and whenever Gosu is ready to do so it'll do. In my cases always instantly or at least so fast that I didn't notice any remarkable delay. 

> Feel free to undo parts of my commit, I've been meaning to finish this today but then I ran out of tinkering time again :(

No problem - the weekends are always too short, but we're in no hurry I guess. At least no one else seems to prioritize any of the open topics.

jlnr:
Ah, another fun complication: Hiding the application with cmd+H also makes the test for `SDL_WINDOW_HIDDEN` return `true`. Of course, we do *not* want to stop the mainloop just because the user has hidden the app for a bit. Oh well...

jlnr:
I think it all works now 🎉

PR #382, opened 2017-03-09
Add (Drag &) Drop callback to Window (Resolves #378)

### LINKED ISSUE #378: Support for drag & drop
As requested here - https://www.reddit.com/r/gosu/comments/5revuf/drop_files_into_gosu_window/

Can be implemented cross-platform using SDL 2: https://wiki.libsdl.org/SDL_DropEvent

Not sure if it will work from within command-line apps on macOS, though? There's no Info.plist to edit :(

Kjarrigan:
There is one strange behavior at the moment. The "last" dragged file doesn't trigger an event. So if you drop one file - nothing happens, but if you drop 5 files you get 4 events (always last file missing). Don't know if this an platform or SDL thing or something in my code is wrong but it also occurs with the "official" SDL example for SDL_DropEvent so please test this on other platforms:

My Setup: Ubuntu 14.04 x86_64 3.19.0-80-generic
And SDL Versio < 2.0.5 (don't know how to check the actual number but SDL_DROPTEXT wasn't available so it has to be older than 2.0.5)

As this can't be automatically tested I added an "interactive" test. just set the env-Variable GOSU_TEST_INTERACTIVE and then call rake test to get it. That's an approach I would want to introduce in PR #360 as well, as most stuff could easily tested by a human (do you see a red line?) but not or only with great effort by an automatic setup.


Gosu/Window.hpp
```diff
@@ -111,8 +111,11 @@ namespace Gosu
         //! in your implementation.
         virtual void button_down(Gosu::Button);
         
-        //! Same as button_down. Called then the user releases a button.
+        //! Same as button_down. Called when the user releases a button.
```
Kjarrigan: Just an typo I found - not related to the actual request but didn't want to make a seperate commit / PR

jlnr:
Also, I'm all in favour of manual tests! In fact, I have a dozen or two "feature tests" floating around: single Ruby files that I built to reproduce bugs or to test new features. All of them require manual interaction, and I've often thought that it'd be cool to have a manual test suite for them.

Gosu/Window.hpp
```diff
@@ -111,8 +111,11 @@ namespace Gosu
...
         virtual void button_up(Gosu::Button) {}
+
+        //! Called when you (drag &) drop a file on the window. Serves the path of the dropped file as argument.
+        virtual void drop(char* path) {}
```
Kjarrigan: Hehe. I actually tried that first but I got conversion errors :cry: But I'll try again needed quite a few runs from scratch to make it work so the error might have been cause by something else.

src/Window.cpp
```diff
@@ -237,13 +243,28 @@ bool Gosu::Window::tick()
...
-            input().feed_sdl_event(&e);
+        switch (e.type) {
+            case (SDL_QUIT): {
+                SDL_HideWindow(shared_window());
+                return false;
+            }
+        #if SDL_VERSION_ATLEAST(2, 0, 5)
+            case (SDL_DROPBEGIN):
+            case (SDL_DROPCOMPLETE):
+            case (SDL_DROPTEXT):
```
Kjarrigan: As I dont have SDL 2.0.5 I couldn't really test what the other events do / return. How I interpret the config is that they all return the filename but are triggered by different things? So I thought I just add all of the SDL_DropEvent types here.

Kjarrigan:
I'll look into the Copy&Paste thing, but would be interested in if this works on your OS X (I don't have any MAC-System to try).

src/Window.cpp
jlnr: I'm on SDL 2.0.5 here and adding these additional event types changes nothing, it works just fine with just the SDL_DROPFILES block.

src/Window.cpp
```diff
@@ -237,13 +243,28 @@ bool Gosu::Window::tick()
...
+        #endif
+            case (SDL_DROPFILE): {
+                char* dropped_filedir = e.drop.file;
```
jlnr: If you change this to `std::string`, please add a null check - sadly, `std::string(nullptr)` crashes instead of constructing an empty string.

test/test_window_interactive.rb
```diff
@@ -0,0 +1,57 @@
...
+  require "gosu" unless defined? Gosu
+
+  class TestWindowInteractive < Minitest::Test
+    class DropWindow < Gosu::Window
+      def initialize
+        super(800,600)
+        self.caption = "(Drag &) Drop Test"
+
+        @font = Gosu::Font.new(15)
+        @files = ["Please drop any number of files and make sure all names appear!"]
```
jlnr: This line will be discarded immediately.

test/test_window_interactive.rb
```diff
@@ -0,0 +1,57 @@
...
+        @files = ["Press Y if everything is ok or N if something went wrong!"]
```
jlnr: Too many exclamation marks :)

test/test_window_interactive.rb
```diff
@@ -0,0 +1,57 @@
...
+          close
+        end
+        if id == Gosu::KB_N
+          puts "Drop-Test failed!"
+          close
+        end
+      end
+    end
+
+    def open_file_or_folder(path)
```
jlnr: I think this is overkill for this simple test class.

jlnr:
Just tried it, it works. It's not perfect, as there's no visual feedback until you release the file (usually the mouse cursor changes), but it does the trick.

jlnr:
I'm still a little torn over this, as I've usually tried to keep Gosu from becoming a GUI library. Then again, it's just one extra method on `Window`, and I feel this callback encourages creative tool building. So I think this is still a good feature 👍 Just needs a little more polish and documentation in `rdoc/gosu.rb`.

jlnr:
It works perfectly on Ubuntu GNOME 16.10, so whatever issue you are seeing in 14.04 must have been fixed.

Kjarrigan:
Rebased and made changes as requested. Not sure about the `SDL_free` part. Added it because it was in the example but you seem to free nothing else explicitly. Simplified the tests with an helper (I tried an similar approach in the audio-enhancement branch and it worked quite well with relatively few line of code)

jlnr:
Ah, good catch. Yes, `SDL_Free` is necessary here (according to the example) - unlike in the rest of Gosu, where C++ automatically manages all memory for us.

jlnr:
Merged, I decided to not even label it "experimental" - it's a super lightweight feature :) Thanks.

PR #383, opened 2017-03-09
TravisCI - Use xvfb to remove the SDL Error and therefore the need to…

Kjarrigan:
… rescue it under linux. And remove some minitest - warnings


Kjarrigan:
Um...thats harder than it should be - maybe its too early :smile: tried this in another PR and worked well, but as it's some more general problem solving thing I wanted to make a seperate Request but, oh well....I hope the 10th run or so is finally green again.

jlnr:
What did I miss, why the 10th run? Is this flaky on Travis (timeouts)? This looks like a fantastic solution, I think I'll just merge it and find out. :)

PR #389, opened 2017-03-26
Add deprecations to Ruby/Gosu (Resolves #367, subset of #370)

### LINKED ISSUE #367: Ruby/Gosu needs deprecations

jlnr:
I still have no idea how to safely deprecate constants and methods that have moved from one scope to another. This PR is a subset #370 that focuses on low-hanging fruit:

- Deprecate the Window constructors, this alone should be a good move for the Gosu ecosystem 👍 
- Deprecate `Window#set_mouse_position` and `Font#draw_rot`, which are always and clearly deprecated. (This is also a good time to remove them from rdoc.)
- Deprecate `Gosu::GOSU_COPYRIGHT_NOTICE` because I really want to deprecate at least one constant, and I (sadly...) doubt many people use this one, especially together with `include Gosu` - so this should be safe enough.

### DIFF
--- a/lib/gosu.rb
+++ b/lib/gosu.rb
@@ -10,12 +10,12 @@ if RUBY_PLATFORM =~ /mswin$|mingw32|mingw64|win32\-|\-win32/
   # encoding (see #385).
   ENV["PATH"] = "#{binary_path.encode ENV["PATH"].encoding};#{ENV["PATH"]}"
   
-  # Add the correct directory 
-  RUBY_VERSION =~ /(\d+.\d+)/
-  $LOAD_PATH.unshift "#{binary_path}/#{$1}"
+  # Add the correct lib directory for the current version of Ruby (major.minor).
+  $LOAD_PATH.unshift File.join(binary_path, RUBY_VERSION[/^\d+.\d+/])
 end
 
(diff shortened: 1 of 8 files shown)


lib/gosu.rb
```diff
@@ -10,12 +10,12 @@
   # encoding (see #385).
   ENV["PATH"] = "#{binary_path.encode ENV["PATH"].encoding};#{ENV["PATH"]}"
   
-  # Add the correct directory 
-  RUBY_VERSION =~ /(\d+.\d+)/
-  $LOAD_PATH.unshift "#{binary_path}/#{$1}"
+  # Add the correct lib directory for the current version of Ruby (major.minor).
+  $LOAD_PATH.unshift File.join(binary_path, RUBY_VERSION[/^\d+.\d+/])
```
jlnr: Should have gone into the last commit on `master`, oh well

PR #392, opened 2017-03-28
Experimental feature: add a simple MessageBox

Kjarrigan:
Despite my comment to solve existing issues first I wanted to open this PR as it is in my pipeline for too long and it basically is just one line of code to call an dialog via SDL...

**Why?**
I saw this while working on the Drag&Drop stuff in its example and thought thats pretty simple and might have some use cases. For example one could show the information provided by Gosu::LICENSE with a simple call of

```ruby
Gosu.simple_dialog("About", Gosu::LICENSES)
```

**Todo**
If this is an accepted feature we might move it in a seperate file. To be hones I tried it forever but just couldn't make it work .... I'm just not into the whole C(++) stuff with headers and includes.

### DIFF
--- a/Gosu/Window.hpp
+++ b/Gosu/Window.hpp
@@ -31,6 +31,8 @@ namespace Gosu
     //! Returns the maximum height (in 'points') that is available for a non-fullscreen Window.
     //! All windows larger than this size will automatically be shrunk to fit.
     unsigned available_height();
+
+    int simple_dialog(std::string caption, std::string message);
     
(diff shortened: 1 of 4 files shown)


jlnr:
This is the first PR where I have to use my veto powers :) A message box cannot be dismissed with the gamepad and looks odd in a game. For quick debugging messages, you can abuse `Window#caption=`. Think SDL_ShowSimpleMessageBox might be useful if Gosu wants to report fatal errors such as SDL_Init failing, but then we can just call the C API directly.

Kjarrigan:
No Problem :smile:

PR #394, opened 2017-03-30
Audio Tests

Kjarrigan:
Decoupled from my Song/Sample-Merge approach I thought I could submit my tests earlier as well as some minor _improvements_.

- Extend the RDoc a bit

- Add `Song.resume`

... as I was quite confused when my copied sample test-case complained that there is no resume for Song. I mean there is a pause, so I would expect an resume even if it just an alias and you actually could now do `Song.resume` without ever pausing...

- Add `rake test_interactive`

...just because I'm lazy and don't want to always set / unset the Trigger-ENV


Kjarrigan:
Forgot to rebase against master :smile:

Kjarrigan:
I should have known...

> ALCmmdevPlayback_open: Device init failed: 0x80070490

Guess I'll look if there is a chance to enable a virtual sound device in the CI's similiar to xfvb for the screen stuff.

jlnr:
I've googled a bit because I was curious as well: https://github.com/favonia/pulse/issues/20

Not sure if that's exactly what we need, and it already looks 10x more complicated than Xvfb... :(

Kjarrigan:
Hm I'll try that - haven't found any better approach so far.

Kjarrigan:
How hard can it be.... I found this for Linux:
https://superuser.com/questions/344760/how-to-create-a-dummy-sound-card-device-in-linux-server
which leads to another error which was solved here
http://askubuntu.com/questions/490711/configure-dummy-sound-device-on-server-with-snd-dummy
which needed more code in Travis:
https://docs.travis-ci.com/user/ci-environment/#Group-membership

Basically use the snd-dummy module from ALSA but you have to be in the audio-group which is all easy in a real environment but not that easy to setup in Travis....

Kjarrigan:
Hm it finally is green on linux BUT the whole `sudo -E su -c ''` hides the output from the tests which I really don't like. This way I have to rely that exited with 0 meant everything went well, but we might have skipped tests or warnings or everything like that which will be suppressed as well.

jlnr:
I wonder if it's really worth the trouble if we cannot compare the output anyway. I thought about recording Gosu's output to a file, but then we'd need to cut out all the silence to account for subtle timing differences etc... :(

Kjarrigan:
Some CI part for the Sounds would be nice to at least check if the params and types/ranges work AS expected and that the ?-methods return true/false correctly but you're probably right. Searching for hours for a somewhat suitable solutions is probably not worth our valuable time which could be used for actually solving Gosu issues and not CI configuration... so make the whole thing an interactive-only test?

jlnr:
Sounds good.

Kjarrigan:
With the help of my coworker who has some experience fighting with Travis it finally works at least for Travis and in the end wasn't that much code to add. But I changed one more thing which you might not want - I extruded the custom parts in subscripts as there were too man if else x line with no linebreaks which were hard to read. So if you don't want that I would at least prefer the multiline formatting as done in the script part.

Kjarrigan:
Finally everything is green again...Sorry for the delay. So much to do and so less time :cry:

jlnr:
I know :) Not sure if I arrange for Gosu weekends before May. Anyway, great that this is green 👍

jlnr:
Hmm, I prefer multi-line blocks in `.travis.yml` to more CI-related files, so I'll standardise on the former.

I'll also veto the `Song#resume` method for now because it's not really part of the audio tests, and I'm happy neither with the old nor with the new interface. I think it's unintuitive that `resume` accepts the `looping` argument again, instead of resuming playback with the same settings as before. The old Song interface is bad too, but I hope it'll be easier to understand if it consists of less methods.

jlnr:
It's a little crazy how many styles of multi-line strings YAML supports: http://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

rdoc/gosu.rb
```diff
@@ -534,8 +534,19 @@ def play_pan(pan=0, volume=1, speed=1, looping=false); end
...
+    # @param [Float]
     attr_writer :volume
+    
+    ##
+    # Set the playback speed.
+    # @param [Float]    
     attr_writer :speed
+    
+    ##
+    # Set the amount of panning. Can be anything from -1.0 (left) to 1.0 (right). 0 is centered.
```
jlnr: Hmmm, not sure if this is still true. At some point (maybe -1.0), the left speaker should be at max volume, and for values even smaller than that, the sound should be harder to hear (less voume). Will investigate

jlnr:
Added some rdoc – about time that I played around with all the `play`/`play_pan` arguments to see how they actually work :) Thanks again for the PR.

Kjarrigan:
:+1: We're on our way toward "full" test coverage :smile:

jlnr:
I guess I should start working on render-to-texture code so that the rest can be tested... ^^"

Kjarrigan:
Yes please :heart: That would solve one of the older requests, improve our test capabilities and might lead to some other interessting approaches. I recently tried some stuff with Gosu.record and Image.subimage and it worked quite well and fast. So I thought this could be used for a splitscreen game but at the moment this didn't work because you can't use subimage on record at the moment :smile: (a minimal gist of the concept I thought about https://gist.github.com/Kjarrigan/116e3284aafa0442f778373db4543f2f)

jlnr:
I think that might be inefficient even with render-to-texture. I've implemented splitscreen in Terava using a combination of rotate/translate/scale, clip_to and Gosu.flush. Works quite nicely :)

https://www.youtube.com/watch?v=HyNmL_PogKE&t=82s

ShFil119:
For openAL environment set: `ALSOFT_DRIVERS=null` is enough.

PR #397, opened 2017-04-03
Add MP3 support with libmpg123 (fixes #393)

### LINKED ISSUE #393: Gosu does not support MP3 files
```ruby
require "gosu"

class Window < Gosu::Window
	def initialize
		super(1000, 1000, false)
		Gosu::Song.new("test.mp3").play
	end
end

Window.new.show
```
Where test.mp3 is a fresh mp3 file from audacity, or a downloaded song.
Produces the following error
```
AL lib: (EE) LoadHrtf: Could not open default-44100.mhr
AL lib: (EE) GetHrtf: Incompatible format: 5.1 Side 44100hz
C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/gosu-0.11.3.1-x64-mingw32/lib/gosu/compat.rb:128:in `initialize': File contains data in an unknown format. (RuntimeError)
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/gosu-0.11.3.1-x64-mingw32/lib/gosu/compat.rb:128:in "initialize"
        from audiotest.rb:6:in "new"
        from audiotest.rb:6:in "initialize"
        from audiotest.rb:10:in "new"
        from audiotest.rb:10:in "<main>"
```
Tried both 64 and 32 bit ruby, both latest gosu and gosu 0.10.7 which i had installed
Audio files and code works on Mac

jlnr:
This library looks like it will be pretty easy to integrate, let's give it a try.

- [x] Add `MPEGFile.hpp`
- [x] Bundle libmpg123.dll on Windows
- [x] Delay-load the library on Windows
- [x] Better logic for deciding whether a file is an MP3 file(?) (let's live with the try-catch hack for now)
- [x] Add mpg123 dependency to `.travis.yml`
- [ ] Update "Getting Started on Linux" page


Kjarrigan:
More stuff to rename in the Sound-Merge-Branch :roll_eyes: 
Don't forget to add an additional line in the LICENSES :smile:

Issue #317, opened 2016-02-20
Implement render-to-image / screenshot functionality


jlnr:
Maybe screenshots can be implemented with a few SDL 2 calls: http://stackoverflow.com/a/22339011

Inspired by this thread: https://www.libgosu.org/cgi-bin/mwf/topic_show.pl?tid=1265

jlnr:
Since Gosu does not use the SDL_Renderer_* family, it will take a bit more code than that :(

djzk:
Another feature I'd like to see :)

jlnr:
I think screenshots are a special case, and if we had `render { ... }` in addition to `record { ... }`, we could implement screenshots *and more* with that.

jlnr:
I can't really find a solution that's less convoluted than this... https://github.com/datenwolf/codesamples/blob/master/samples/OpenGL/minimalfbo/minimalfbo.c

Juksefantomet:
Might this be something of use? i might be off, and i see that it's mainly for windows platform, but if it's on the road in the right direction anything helps right? 👍 

https://github.com/jarmo/win32screenshot

jlnr:
Also check out this approach which uses the Ruby OpenGL libraries: https://www.reddit.com/r/gosu/comments/6wjjsr/saving_rendered_image/

Has been posted just three days ago :)

PR #402, opened 2017-05-04
Add Window#background_color

Kjarrigan:
While I looked through the code to remember how the rendering stuff works to figure out if [this](https://wiki.delphigl.com/index.php/Screenshot) would work as an easy screenshot command #317 I remembered an old wish of mine to change the window background color. And there right in the code was already a _clear_with_color_ variable - at least until I pulled the latest master. So this PR reverts some of the latest changes and provide a function to actually manipulate the _clear_with_color_-value.

I guess the rest is self-explaining.

### DIFF
--- a/Gosu/Graphics.hpp
+++ b/Gosu/Graphics.hpp
@@ -39,7 +39,7 @@ namespace Gosu
 
         //! Prepares the graphics object for drawing and then runs the rendering code in f.
         //! Nothing must be drawn outside of frame() and record().
-        void frame(const std::function<void ()>& f);
+        void frame(Gosu::Color clear_with_color, const std::function<void ()>& f);
         
         //! Flushes the Z queue to the screen and starts a new one.
         //! This can be useful to separate the Z queues of two parts of the game, e.g. the two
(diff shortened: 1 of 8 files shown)


jlnr:
I like the implementation (although RED is not a good default :P) but I don't think the feature is worth the extra callback. If the user wants a different background colour, they can fix it with a single `draw_rect` call. That's also why I've removed the argument to `frame`. Sorry, I'm going to defend Gosu's small interface here again :)

Kjarrigan:
Yeah the red was for testing if it works and somehow survived the cleanup :smile: No problem. I have some more ideas so plenty of chances to get something merged :grin:

akicho8:
I recently started using Gosu. However, I am having trouble figuring out how to change the background color, which is why I ended up here. I know that you can overwrite it by calling draw_rect, but isn't that a waste of processing if I could specify the color from the beginning?

jlnr:
Using `draw_rect` should not have a measurable impact on rendering. You can render tens of thousands of rectangles per frame even using old hardware. Also, I suspect that most game projects that are approaching completion will use background textures anyway, not solid colors.

akicho8:
Thank you for your prompt response!
I wasn't aware that draw_rect was that fast, even for filling the entire screen. I understand that the background would typically be texturized in most modern games, but my purpose for using Gosu is to simulate physical movements. Therefore, simply clearing the screen with a single color in the fastest way possible is more than enough for my needs. For now, I'm going to proceed with the method of using draw_rect.

PR #421, opened 2017-11-20
Audio Refactoring

jlnr:
The massive Audio.cpp file was a bit of an embarrassment, and got into the way of me understanding the two existing audio refactoring branches. So I've started a third branch! (Sorry @Kjarrigan 😅)

This branch aims to:
* Rename `SampleInstance` to `Channel`
* Add default constructors to `Sample`, but only in C++, like `Image`. `Song` does not receive a default constructor because it is non-copyable (not sure if this logic is sound)
* Refactor the awkward ALChannelManagement class into an encapsulated little thing
* Improve everything I could find on the way

It should not:
* Break Ruby backward compatibility
* Make anything thread-safe that wasn't thread-safe before (it's quite tempting!)
* Merge Song and Sample (I'll look into this after merging this PR)

More ideas:
* [x] `ChannelManagement.{hpp,cpp}` should probably be called `AudioImpl.{hpp,cpp}` to be more consistent with `GraphicsImpl.hpp`
* [x] Fix Travis on this branch, where I can comfortably force-push, and then cherry-pick this onto the fix before merging.


Kjarrigan:
Hi @jlnr ,

no problem. I have to apologize, that I never finished the audio branch (and some other stuff :cry:) Since my second daughters birth I hadn't much time. I try to follow the issues and commits but sadly can't do more atm. I hope in the not-to-far-future I can return to more regular contributions.

:+1: for your work

jlnr:
@Kjarrigan No need to apologize! :) There are plenty of useful commits in your branches, thanks for your work. I've been almost completely idle for the last months as well.

pmer:
Adding a default constructor for Song sounds good to me. 👌

lib/gosu/patches.rb
```diff
@@ -45,6 +45,11 @@ def eql?(other)
   end
 end
 
+module Gosu
+  # Backwards compatibility: Channel was called SampleInstance before Gosu 0.13.0.
+  SampleInstance = Channel
```
Kjarrigan: Just a minor nitpick - wouldn't it be better to move it to compat and use deprecate_const like we did it with the GOSU_COPYRIGHT_NOTICE?  And even if you don't like the deprecation warning it would be more consistent in the compat file because 

> # This whole file is about backward compatibility.

Otherwise nice job :smile:

jlnr: Oh wow, you are right. Thanks for pointing that out! Will fix it ASAP.

PR #422, opened 2017-11-26
Use stb_truetype for all text rendering

### LINKED ISSUE #359: Font.cpp: Use UCS-4 on all platforms, maybe precompose?
Gosu::Font assumes that every `wchar_t` in a string should be rendered as a single character. That leads to inconsistencies between UNIX (`wchar_t` = 32 bit) and Windows (`wchar_t` = 16 bit).

The correct fix would be to split the string into [grapheme clusters](http://unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries), but I don't know if that is possible with Gosu's current dependencies (Win32 API, macOS API, iconv). It is also frustrating that the definition of a grapheme cluster changes every year. [This page looks completely different on macOS 10.11 and 10.12.](http://emojipedia.org/emoji-zwj-sequences/)

The next best thing is to convert the UTF-8 string into [precomposed code points](https://en.wikipedia.org/wiki/Precomposed_character). This avoids Ä being rendered as A¨ and also ensures that UTF-16 surrogates are not rendered as separate characters, as is the case on Windows right now.

And if precomposing is impossible without external dependencies, Gosu should at least use UCS-4 on Windows for consistency with UNIX.

jlnr:
Goals:

* [x] Add tests for text rendering – this is possible without render-to-image functionality, we can use `Image.from_text(...).save(...)`
* [x] Ditch macOS-/iOS-specific code by using [stb_truetype](https://github.com/nothings/stb/blob/master/stb_truetype.h)
* [x] Ditch Windows-specific code by using [stb_truetype](https://github.com/nothings/stb/blob/master/stb_truetype.h)
* [x] Ditch Linux-specific code by using [stb_truetype](https://github.com/nothings/stb/blob/master/stb_truetype.h)
* [ ] Make `Font` handle Unicode more sensibly by bundling & using [utf8proc](https://github.com/JuliaLang/utf8proc) (fixes #359)
* [ ] Specifically, replace `Font#[]=` by a block passed to the constructor, for example: `Font.new(...) { |grapheme| Image.new("glyphs/#{grapheme}.png") }`. The "default block" just happens to render the grapheme cluster using `Image.from_text`. This will also be the loophole to test grapheme cluster parsing—the block will be called once per cluster.
* [ ] Refactor markup handling, it's inefficient and probably buggy
* [ ] Make `register_entity` available in Ruby (it wasn't?!)
* [ ] Add `Font#draw_markup` and `Image.from_markup` as suggested in #304, but make these aliases for the existing methods for now; Gosu 1.0.0 will then introduce a difference


jlnr:
Idea: Gosu currently uses OS APIs to locate and render fonts when a non-filename is passed as the `font:` option. If we can lookup TTF files by font name, Gosu could use `stb_truetype` even for system fonts. On Linux, that probably involves using the `fontconfig` API: https://github.com/xbmc/fontconfig/blob/master/fc-match/fc-match.c

jlnr:
When I run this:

`ruby -rgosu -e 'Gosu::Image.from_text("Was <b>geht <i>alder</i></b>", 40, font: "Arial").save("test.png")'`

...the result does not look satisfactory on Linux:

test

I need to check if this is caused by a terrible version of arial.ttf being loaded (shouldn't be?) or if my rendering code is just bad.

jlnr:
This PR has been around long enough, I'll move the remaining bullet points into one or two separate branches.

PR #423, opened 2017-12-01
Get Gosu to work in Visual Studio 2015 x64

oli-obk:


jlnr:
Hmmm. I am using VS 2015 all the time, and it was my understanding that WIN32 is merely the name of the C-based API, not necessarily tied to the "bittiness" of the build. Does VS 2015 not set this `#define` for new projects?
Is this fix necessary for building from source, or for using the pre-built `.lib` files?

jlnr:
I am probably just going to merge this anyway (what could go wrong?), but I haven't played around with C++ Gosu on Windows for a while and I'm curious :)

oli-obk:
> it was my understanding that WIN32 is merely the name of the C-based API, not necessarily tied to the "bittiness" of the build

That's what I thought, too.

You can try undoing this change in https://github.com/oli-obk/dhbw-objektorientierung and see if the project compiles on your machine. As you can see in https://github.com/oli-obk/dhbw-objektorientierung/commits/oli-obk-patch-1 , it fails on travis if I undo the change.

pmer:
StackOverflow suggests that the _WIN32 macro is defined on both 32 and 64-bit builds.

https://stackoverflow.com/questions/17380340/win32-preprocessor-definition-in-64bit-windows-platform

Hmm, it's confusing that WIN32 and _WIN32 are different.

jlnr:
Well, it clearly doesn't hurt, so here comes the merge!

@oli-obk Are you using the pre-compiled .lib files that I push to GitHub? I've been wondering if I should continue to do so, or if C++ hackers are just going to compile them on their own anyway :)

oli-obk:
@jlnr yea we're using the pre-compiled .lib files. If I can get Gosu to compile as a dependency without any additional installations, I'll move away from that, because then I can put  this repository as a submodule into my repository and future updates become just a git pull instead of getting the zip, extracting, and making a new commit.

jlnr:
@oli-obk Hmmm, no idea how that could work as long as Gosu depends on SDL 2. You could add this [SDL git mirror](https://github.com/spurious/SDL-mirror) as another submodule, but I'm not sure how you would build it for Gosu to link against.

Anyway, great to hear that the `.lib` files actually work! I'll just keep uploading them for now, it's only a couple of extra clicks anyway.

PR #429, opened 2018-02-07
Refactor Gosu::Font

### LINKED ISSUE #359: Font.cpp: Use UCS-4 on all platforms, maybe precompose?

jlnr:
This PR aims to implement the following:

* [ ] Replace `Font#[]=` by a block passed to the constructor, for example: `Font.new(...) { |grapheme| Image.new("glyphs/#{grapheme}.png") }`.
  The "default block" just happens to render the grapheme cluster using Image.from_text.
* [ ] Make Font handle Unicode more sensibly by bundling & using [utf8proc](https://github.com/JuliaLang/utf8proc) (fixes #359)
* [ ] Decide what to do with font's default flags. Right now it is always bold - I think it shouldn't be, but Font.new should accept `bold: true, italic: true` as an argument (breaking change!)
* [x] Add unit tests

Since the markup handling is still based on `std::wstring`, not UTF-8, it will need to be refactored as well:

* [x] Refactor markup handling, it's inefficient and probably buggy
* [x] Remove &text_entity; functionality (which was never exposed to Ruby anyway)
* [ ] Add `Font#draw_markup` and `Image.from_markup` as suggested in #304, and remove markup handling from the current methods (breaking change)
* [ ] Add unit tests

Also:

* [x] Make sure CJK text still has moderately sane line breaks even in the absence of whitespace

Hrmph:

* [ ] Refactor new TTF code to add font fallback logic, otherwise CJK text can't be rendered using `Gosu.default_font_name` :/


jlnr:
Merging current progress to allow other PRs to be merged first – and also my work on https://github.com/gosu/motion-gosu/issues/6

PR #431, opened 2018-03-05
Implement Window.to_bitmap - Resolves #317

### LINKED ISSUE #317: Implement render-to-image / screenshot functionality

Kjarrigan:
Hi there,

finally had some tinkering time again and started looking at the screenshot again (via glReadPixels). Gave me a big headache but finally I think I'm kinda on the right way. Although it doens't work so far (I guess I messed up the colored channels/types) I want to share my general approach.

First things first - the naming... Tried quite a few things but in the end I always ended up creating a Gosu::Bitmap one way or the other and then saved it to disc, but then I thought - why should I do this - getting the bitmap is the hard part - let the user do the rest. With an acutal Bitmap (or Image in Ruby) object your far more flexible to do stuff. You could instantly reuse the image/subimage/resize it e.g. as Thumbnail for the Savegame-Slot without saving it to disc and then reload it. Or do some fancy filter effects. The saving can still easily be done by calling window.to_image.save("screenshot.bmp").

So hoping for feedback! I'm done for today but hope to get back to work tomorrow.

PS: The ASCII Art of the triangle from my simple test currently looks like this...

```
######################
#@                   #
#                    #
#        .iiii..iii.i#
#...i        iii     #
#        iii..ii..   #
#                    #
#  i.iii..ii..iii    #
#                    #
#.ii..iii:ii..       #
#          iiiii     #
#      @ iii..iiii.. #
#                    #
#  @ iii..iiii..i    #
#                  @ #
#iii..iiii..iii.i...i#
#              i.iii.#
#.iiii..iii.i...i    #
#    iii           @ #
#iii..iiii..         #
#              @ iii.#
######################
```

PPS: I also tried to implement the Macro.to_bitmap function (that's how I came up with Window.to_bitmap in the first place. But always got "No rendering queue". If the window part works I might do a follow up and try to fill that empty space too.)


Kjarrigan:
Update - not its working! Some small issues though. 

1) Of course you can only take a screenshot if the window is really drawn (so I needed two window.ticks and a short sleep to reliable pass the tests on my local pc)

2) I had to use the "expensive" set_pixel function, but I felt no delay in my tests so it might be no a issue at all. glReadPixel itself is stated as expensive operation, so....

3) The tests are really simple. Would be nice if someone could try this in his actual game with lots of drawn elements and updates.

The new bitmap operator is now obselete. I'll remove the commit in the final rebase.

Gosu/Bitmap.hpp
```diff
@@ -31,6 +31,11 @@ namespace Gosu
         : w(w), h(h), pixels(w * h, c)
         {
         }
+
+        Bitmap(unsigned w, unsigned h, std::vector<Color> pixels)
+        : w(w), h(h), pixels(pixels)
```
jlnr: Should call std::move on pixels to prevent a copy

ext/gosu/gosu.i
```diff
@@ -1014,6 +1014,12 @@ namespace Gosu
     {
         $self->Gosu::Window::close();
     }
+
+    Gosu::Image to_image()
```
jlnr: Huh, does that work? I think I'm returning `Gosu::Image*` in other places. Would be nice if it could all use the pointerless variant

ext/gosu/gosu.i
```diff
@@ -1014,6 +1014,12 @@ namespace Gosu
...
+    {
+        Gosu::Image image($self->to_bitmap());
```
jlnr: I don't think the temporary variable adds a lot here, better to just return Gosu::Image(...)

src/RubyGosu.cxx
```diff
@@ -11351,6 +11418,7 @@ _wrap_translate(int argc, VALUE *argv, VALUE self) {
 
 /* -------- TYPE CONVERSION AND EQUIVALENCE RULES (BEGIN) -------- */
 
+static swig_type_info _swigt__p_Gosu__Bitmap = {"_p_Gosu__Bitmap", "Gosu::Bitmap *", 0, 0, (void*)0, 0};
```
jlnr: If SWIG wants to add the Bitmap type, you might have forgotten to %ignore Window::to_bitmap somewhere - Bitmap should be invisible to Ruby

Gosu/Bitmap.hpp
jlnr: But it also doesn't look like you're using this new constructor :)

src/Window.cpp
```diff
@@ -331,6 +333,29 @@ void Gosu::Window::button_down(Button button)
     }
 }
 
+Gosu::Bitmap Gosu::Window::to_bitmap()
+{
+    ensure_current_context();
+
+    std::vector<Color> pixels(width()*height());
+    std::vector<Color::Channel> channels(width()*height()*4);
```
jlnr: From a performance POV it would be nicer to create the Bitmap first, then `glReadPixels` into the bitmap (it's designed to be compatible with GL_RGBA, see https://github.com/gosu/gosu/blob/master/src/TexChunk.cpp#L118), and then fix the bitmap's orientation by swapping each row with memcpy and a temporary buffer or plain XOR.

Gosu/Bitmap.hpp
jlnr: Ah, just saw your comment -- n/m

src/Window.cpp
Kjarrigan: I tried memcpy in my first few test and always got segmentation faults but might give it a second shot now that I know that glReadPixels is actually working and returning the right values.

src/RubyGosu.cxx
Kjarrigan: Yes - missed that ;-)

ext/gosu/gosu.i
Kjarrigan: Well - I got no errors, so I guess it works?

Gosu/Bitmap.hpp
Kjarrigan: Nope. Not any more. Tried different things with memcpy the data directly, which didn't worked. Then I tried this solution by assigning the pixels vector directly instead of using set_pixel, which didn't worked either...

Kjarrigan:
@jlnr Thanks for your response :+1: The current implementation acutally works although as stated might not be the most performant. But I'll try the memcpy approach again, now that I figured out the rest.

jlnr:
Hey, great to see another PR :) I agree that window.to_image.save(...) is the best interface for taking screenshots. I'm just a little nervous because you said it required two ticks *and* sleeping for a bit. My only real game doesn't compile right now, but I can try it with some other Ruby/Gosu stuff in the next days.

I think we should merge this branch ASAP just to be able to use it in tests, I could need some visual verification in my Font branch. If we can add a guideline for reliably taking screenshots (without sleeping :)) we can also advertise this as an experimental feature.

I wonder if we can also reuse this interface if we ever re-implement Window#to_image and Image.render using a proper offscreen buffer. Would it be too surprising if Window#to_image would just be shorthand for `render(self.width, self.height) { self.draw }` later? If you use `rand()` in `#draw` then it wouldn't save exactly what's shown on screen, but it would work in edge cases e.g. before the first frame is rendered.

Kjarrigan:
Well the sleep is actually only necessary in the test. If you start your game by window.show everything is rendered but at least at my pc if I call window.tick it didn't always shows a correctly rendered window. It flickered and was gone and so the glReadPixels returned only nonsense.

> I think we should merge this branch ASAP just to be able to use it in tests, I could need some visual verification in my Font branch. If we can add a guideline for reliably taking screenshots (without sleeping :)) we can also advertise this as an experimental feature.

FullACK. I guess thats one of the most requested features so I hope we get some feedback and yeah - the testsuite could grow quite a bit if this finally works.

> I wonder if we can also reuse this interface if we ever re-implement Window#to_image and Image.render using a proper offscreen buffer.

I hope so.

src/Window.cpp
Kjarrigan: Looked again at TexChunk and the docu. The problem is, that glReadPixels doesn't return a single uint32 containing the 4 channels but instead giving me 4 single bytes which i have to combine to a color first before assigning it to the bitmap. Would be cool to have something like Array.pack in Ruby to just squash them together or better making glReadPixels actually return the pixels as one value.

Kjarrigan:
Ok - now at least the testuite under Linux works (its still red because of some Font releated problems). Windows still has the missing opengl constant and mac can't compile at all, but I don't think it is related to this PR.

Kjarrigan:
Its really annoying.... the screenshot should help us test stuff we otherwise can't but its nearly impossble to get platform independet binary identical screenshots....the rendering on every system has some degree of choice how to actually render stuff. The gradients, diagonal lines argh.....The differences are usually really small - as human you can clearly detect it as "they are basically the same" but how to check for this on an automated test.

The colors in the gradients were all just a few bits off. The diagonals are just a (few) pixels off as well. See the attached image. (at appveyor the diagonal is "moved" 1px to the right)

@jlnr - now sure how to deal with that. For the pure screenshot function I would say its perfectly fine. But for the planned tests of all kinds of stuff this has to work somehow. Maybe I should look for an image diff tool that can handle differences up to a given delta...

dummy

Kjarrigan:
Yeah!!!!! @jlnr I finally did it....not perfect and up for review but it looks rather promising after all those failed builds....

Basically the screenshots works 100% perfect - WYSIWYG, but the tests have to ensure that what we mean is what we get and "meaning" has some room for interpretation depending on your platform. So the image diff with an threshold of 90% should compensate that.

The only thing that still bothers me is the double-tick-sleep thingy. Even the 1s is sometimes not enough for me if my system has high cpu-load. I can tell if the test fails beforehand because I can see the window not correctly rendering. Maybe we can just remove the tick-sleep as the testing system should have no ressource stealing side jobs and only focus on the one task we're giving it.

jlnr:
(y) I think this helper makes a lot of sense for screenshot tests. I was thinking that we could maybe get away with only tests that look the same on every computer, like drawing single-color quads that are perfectly aligned (`draw_rect`), lines that aren't diagonal etc., but we can't even test `draw_triangle` that way :/

Another approach I've seen is to color some parts of the screenshot pink (or another "magic" color), to indicate that they should be ignored during comparisons. Maybe we can mix and match both approaches later, or maybe it won't matter :)

jlnr:
Two more thoughts about `Window#to_image` for end users:

* I've read that glReadPixels on the primary renderbuffer is not quite reliable, and on some systems you could even end up reading data from overlapping windows. That sounds a little scary...I think I'm going to look into implementing a proper `render` method based on your current glReadPixels approach and an off-screen framebuffer. That would make the method more stable for end users *and* fix the tick-tick-sleep pattern in the tests.
* If we use offscreen rendering for taking screenshots (because it seems that this is considered the cleaner approach), is `Window#screenshot` maybe the more forward-looking name because we are not *really* copying window contents (in the future)?

lib/gosu/patches.rb
```diff
@@ -57,6 +57,31 @@ def tick
   end
 end
 
+class Gosu::Image
+  def ==(img)
```
jlnr: I don't think this should go into the public Gosu interface as `to_blob` will throw an exception for macros; and even if `to_blob` works on macros in the future (based on `render{}`), I'm not sure if a macro should be equal to an image just because it renders to the same pixel data at a given resolution.
Explicitly comparing `to_blob` avoids all of these issues.

lib/gosu/patches.rb
```diff
@@ -57,6 +57,31 @@ def tick
...
+    self.to_blob == img.to_blob
+  end
+
+  # Checks if two images are similar on a really basic level (check the difference of each channel)
+  def similar_to?(img, treshold=0.90)
```
jlnr: Typo: *`threshold`

I think `similar?(img)`  would be more in line with Ruby's `equal?`. And I don't think that this method is useful outside of tests, so it can go into the `test_helper.rb` file for now.

lib/gosu/patches.rb
jlnr: (And I realize that this means we should also kick `Image#==` out of `test_helper.rb`)

rdoc/gosu.rb
```diff
@@ -771,6 +771,18 @@ def show; end
     # @return [true, false] whether the {Window} should still be shown after this tick
     def tick; end
 
+    ##
+    # EXPERIMENTAL - MAY DISAPPEAR WITHOUT WARNING.
+    #
+    # Creates a Gosu::Image from the current framebuffer (aka Screenshot) which can then be used for various
+    # things most commonly save it to a file.
+    #
+    # Please provide us with feedback in the forum (https://www.libgosu.org/) or at github (https://github.com/gosu/gosu)
```
jlnr: * or on GitHub

src/Window.cpp
```diff
@@ -331,6 +332,34 @@ void Gosu::Window::button_down(Button button)
     }
 }
 
+Gosu::Bitmap Gosu::Window::to_bitmap()
+{
+    ensure_current_context();
+
+    Gosu::Bitmap bitmap(width(), height());
+#ifdef GL_UNSIGNED_INT_8_8_8_8_REV
+    std::vector<std::uint32_t> pixels(width()*height());
```
jlnr: +space around `*` and `<` in this method

Kjarrigan:
> I've read that glReadPixels on the primary renderbuffer is not quite reliable, and on some systems you could even end up reading data from overlapping windows. That sounds a little scary...

I think that might be the case only for overlapping windows of the same gosu-process but offscreen-rendering would definitely be the best approach. Wouldn't that also mean we could completely remove the whole Macro-Special code? Because we could render to an offscreen buffer and then read it back as "true" image so automagically gl-Blocks work and `draw_line` in `Window#draw `would be truly equal to a call in `Gosu#record` .

> Window#screenshot

.. is fine by me. Went a bit back and forth with the naming and just sticked to `to_image` as there was already an to_bitmap. Should the C++ interface still be `to_bitmap` or named screenshot as well?

> Another approach I've seen is to color some parts of the screenshot pink (or another "magic" color), to indicate that they should be ignored during comparisons. Maybe we can mix and match both approaches later, or maybe it won't matter :)

Never seen this approach but sounds reasonable to me. I'll think about this and the color could be another parameter so the bitmappers could use their pink and our default window could just use its black...

Pushed the chances and hopefully finally ready for merge :smile: Took me far longer than I thought, but I guess the brain mangling part is done. Now its just the cleanup :tada:

ext/gosu/gosu.i
```diff
@@ -1016,7 +1016,7 @@ namespace Gosu
         $self->Gosu::Window::close();
     }
 
-    Gosu::Image to_image()
+    Gosu::Image screenshot()
```
jlnr: 👍

jlnr:
Thanks again for the PR! Can't wait to write more tests :)

jlnr:
> Wouldn't that also mean we could completely remove the whole Macro-Special code? Because we could render to an offscreen buffer and then read it back as "true" image so automagically gl-Blocks work and draw_line in Window#drawwould be truly equal to a call in Gosu#record .

Macros are still useful if you want to render a large tiled map that would consume too much space if you used render-to-texture, and macros are also a bit like vector graphics because you can rotate and scale them without pixelation artifacts. It just means that the difference between `render` and `record` has to be much better documented (like so much in Gosu :/).

PR #432, opened 2018-03-06
[WIP] Add Song#finished callback - Resolves #188

### LINKED ISSUE #188: Callbacks for sample/song playback completion
The callbacks could be called at, say, the end of the event loop. I see it being useful for automatically playing the next song in a playlist, or maybe for songs that have a unique intro or fade-in followed by a cleanly looping part.

A hypothetical example:

``` ruby
def play_looping_song_with_intro
  intro = Song.new(window, 'intro_that_plays_once.ogg')
  song = Song.new(window, 'looping_part.ogg')

  intro.play { song.play(true) }
end
```

Kjarrigan:
Didn't work at the moment because I got some problems with SWIG (again). I tried to remember what I did in the Window#drop callback but couldn't. Swig doesn't register the finished with an upcall wrapper so the overloaded ruby function never is called. Any ideas?

Some general thoughts:

- I added it only to song because there you basically already checked for finished and I can't see any use-case in a Sample right now.
- Don't know lambdas good enough - do you think providing an constructor like Song(filename, callback-lamdba) would be helpful for the C++ users or would they instinctively override finished directly?
- We could theoretically remove the looping logic and define it as finished callback instead, as its basically the same idea: Play a / the same song as soon as the old one is finished.

### DIFF
--- a/src/RubyGosu.cxx
+++ b/src/RubyGosu.cxx
@@ -8377,6 +8377,34 @@ fail:
 
 
 SWIGINTERN VALUE
+_wrap_Song_finished(int argc, VALUE *argv, VALUE self) {
+  Gosu::Song *arg1 = (Gosu::Song *) 0 ;
+  void *argp1 = 0 ;
+  int res1 = 0 ;
+  
+  if ((argc < 0) || (argc > 0)) {
+    rb_raise(rb_eArgError, "wrong # of arguments(%d for 0)",argc); SWIG_fail;
+  }
+  res1 = SWIG_ConvertPtr(self, &argp1,SWIGTYPE_p_Gosu__Song, 0 |  0 );
+  if (!SWIG_IsOK(res1)) {
+    SWIG_exception_fail(SWIG_ArgError(res1), Ruby_Format_TypeError( "", "Gosu::Song *","finished", 1, self )); 
+  }
+  arg1 = reinterpret_cast< Gosu::Song * >(argp1);
+  {
+    try {
+      (arg1)->finished();
+    }
+    catch (const std::exception& e) {
+      SWIG_exception(SWIG_RuntimeError, e.what());
+    }
+  }
+  return Qnil;
+fail:
+  return Qnil;
+}
+
+
+SWIGINTERN VALUE
 _wrap_Song_update(int argc, VALUE *argv, VALUE self) {
   if ((argc < 0) || (argc > 0)) {
     rb_raise(rb_eArgError, "wrong # of arguments(%d for 0)",argc); SWIG_fail;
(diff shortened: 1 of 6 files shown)


src/RubyGosu.cxx
```diff
@@ -8376,6 +8376,34 @@ _wrap_Song_volumee___(int argc, VALUE *argv, VALUE self) {
 }
 
 
+SWIGINTERN VALUE
```
Kjarrigan: This should include an upcall-branch to actually call the ruby function if available but don't know how to trigger it.

test/test_audio.rb
```diff
@@ -114,4 +114,26 @@ def test_song
...
+    song1 = Gosu::Song.new(media_path("0614.ogg"))
+    song2 = Gosu::Song.new(media_path("0830.ogg"))
+
+    song1.play(false) { song2.play }
+    assert_equal song1, Gosu::Song.current_song
+
+    assert song1.playing?
+    refute song2.playing?
+
+    (20 * 1000/win.update_interval).ceil.times{ win.tick }
```
Kjarrigan: Maybe I should look for a shorter first sound-clip as 20s are quite some time...

jlnr:
Thanks for the PR. Random thoughts:
* The SWIG feature to look for is called "director". Right now it's only enabled for TextInput and Window, and it does the automatic upcall.
* However, I think I prefer a constructor block because that makes it a bit easier to bridge it to Ruby, and because I think it would be more practical if you want to refer to non-global variables from within the handler.
  I'm adding a c'tor block to Gosu::Font in #429, so Song wouldn't stick out as the only class with a c'tor block after that. (We might need to explicitly hold a reference to the block from the Ruby side though - Gosu::gl does that for its blocks to prevent the GC from freeing them before Gosu has called them. I'm not sure why the Ruby GC can't look "into" std::function, and maybe this has changed.)
* I'm not sure that this feature is super useful on Gosu::Song when any part of the code can currently poll `Gosu::Song.current.playing?` and start a new song. The reason that the looping parameter exists as a special case is that Gosu can seamlessly loop the current song/sample, whereas a callback block will introduce short but audible gaps if it executes on the main thread. I'm afraid we can't drop it that easily :/

I'm not sure if it makes sense to merge this before merging the Sample and Song classes, because it increases the diff between both classes, and it still doesn't allow for relatively basic features like cross-fading between songs. I think I'd prefer to merge both classes into a more flexible Sound class first, that works more like Image (read/write from raw data, get current position, get length of song in fractional seconds etc.)

jlnr:
Another idea that I just had was to add a block argument to `play` that is fired when the sound ends. However that would break Song's immutability, because every `play` call would overwrite the previous call. It would work for Sample, but again I'm not sure if this would be too heavy-weight as a general solution.

Kjarrigan:
Hi - on the way home I thought about my approach as well and had some other ideas (than the current approach) as well and I guess your right about merging Song and Sample first. Thats what made me look into this PR in the first place. Its been a long time since I last looked into the Audio Implementation and so I wanted to _slowly_ get back into it by resolving a "low hanging fruit" (at least it looked like one while reading the code).

jlnr:
One more random thought: Overriding a method works for Gosu::Song right now because this type cannot be copied or assigned to. Overriding a method on Gosu::Sample (or a unified Gosu::Sound) would be awkward as users would have to actively [slicing](https://stackoverflow.com/a/274634).

I do wonder if overridable virtual methods wouldn't be a good choice for Gosu::Input though. The public `std::function<>` members definitely stick out. This is an internal C++ thingie though.

Kjarrigan:
As discussed - lets merge Sample and Song first and then I might reopen this or maybe it could even be solved while doing the merge.

PR #437, opened 2018-03-09
Resizable Window

cyberarm:
Made Gosu's Window resizable via the window manager, you'll probably find a better and more compatible way to do this.
(Only tested on Windows 10, Ruby 2.5)

P.S.
Atom decided to trim some white space...

### DIFF
--- a/Gosu/Window.hpp
+++ b/Gosu/Window.hpp
@@ -41,7 +41,7 @@ namespace Gosu
         unsigned height() const;
         bool fullscreen() const;
         bool resizable() const;
-        void resize(unsigned width, unsigned height, bool fullscreen, bool resizing = false);
+        void resize(unsigned width, unsigned height, bool fullscreen);
 
         double update_interval() const;
         void set_update_interval(double update_interval);
--- a/Gosu/Window.hpp
+++ b/Gosu/Window.hpp
@@ -34,14 +34,14 @@ namespace Gosu
         //! \param update_interval Interval in milliseconds between two calls to the update member
         //! function.
         Window(unsigned width, unsigned height, bool fullscreen = false,
-            double update_interval = 16.666666);
+            double update_interval = 16.666666, bool resizable = false);
         virtual ~Window();
 
         unsigned width() const;
         unsigned height() const;
         bool fullscreen() const;
         void resize(unsigned width, unsigned height, bool fullscreen);
-        
+
         double update_interval() const;
         void set_update_interval(double update_interval);
 
--- a/Gosu/Window.hpp
+++ b/Gosu/Window.hpp
@@ -40,7 +40,8 @@ namespace Gosu
         unsigned width() const;
         unsigned height() const;
         bool fullscreen() const;
-        void resize(unsigned width, unsigned height, bool fullscreen);
+        bool resizable() const;
+        void resize(unsigned width, unsigned height, bool fullscreen, bool resizing = false);
 
         double update_interval() const;
         void set_update_interval(double update_interval);
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -187,7 +187,7 @@ void Gosu::Window::resize(unsigned width, unsigned height, bool fullscreen, bool
     
     SDL_SetWindowFullscreen(shared_window(), fullscreen ? SDL_WINDOW_FULLSCREEN_DESKTOP : 0);
     if (!resizing) { // Don't set window size if window was resized by user
-    SDL_SetWindowSize(shared_window(), actual_width, actual_height);
+        SDL_SetWindowSize(shared_window(), actual_width, actual_height);
     }
     
 #if SDL_VERSION_ATLEAST(2, 0, 1)
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -177,17 +177,18 @@ void Gosu::Window::resize(unsigned width, unsigned height, bool fullscreen, bool
         double max_width  = Gosu::available_width(this);
         double max_height = Gosu::available_height(this);
         
-
+        // Don't use scaling if window was resized by user
         if (!resizing && (width > max_width || height > max_height)) {
             scale_factor = min(max_width / width, max_height / height);
             actual_width  = width  * scale_factor;
             actual_height = height * scale_factor;
         }
-        printf("screen size available %fx%f, actual size %dx%d\n", max_width, max_height, actual_width, actual_height);
     }
     
     SDL_SetWindowFullscreen(shared_window(), fullscreen ? SDL_WINDOW_FULLSCREEN_DESKTOP : 0);
+    if (!resizing) { // Don't set window size if window was resized by user
     SDL_SetWindowSize(shared_window(), actual_width, actual_height);
+    }
     
 #if SDL_VERSION_ATLEAST(2, 0, 1)
     SDL_GL_GetDrawableSize(shared_window(), &actual_width, &actual_height);
@@ -208,9 +209,6 @@ void Gosu::Window::resize(unsigned width, unsigned height, bool fullscreen, bool
     }
     pimpl->input->set_mouse_factors(1 / scale_factor, 1 / scale_factor,
                                     black_bar_width, black_bar_height);
-
-    printf("Resizing to %dx%d from target %dx%d, was %dx%d\n", actual_width, actual_height, width, height, pimpl->graphics->width(), pimpl->graphics->height());
-    printf("Scale Factor: %f, black bar: %fx%f\n\n", scale_factor, black_bar_width, black_bar_height);
 }
 
 double Gosu::Window::update_interval() const
@@ -288,7 +286,7 @@ bool Gosu::Window::tick()
                     #endif
 
                     case SDL_WINDOWEVENT_SIZE_CHANGED: {
-                        if (pimpl->resizable && width() != e.window.data1 || height() != e.window.data2) {
+                        if (pimpl->resizable && (width() != e.window.data1 || height() != e.window.data2)) {
                             resize(e.window.data1, e.window.data2, fullscreen(), true);
                         }
                         break;
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -100,9 +100,9 @@ struct Gosu::Window::Impl
 Gosu::Window::Window(unsigned width, unsigned height, bool fullscreen, double update_interval, bool resizable)
 : pimpl(new Impl)
 {
-    #if SDL_VERSION_ATLEAST(2, 0, 5)
-        SDL_SetWindowResizable(shared_window(), (SDL_bool)resizable);
-    #endif
+#if SDL_VERSION_ATLEAST(2, 0, 5)
+    SDL_SetWindowResizable(shared_window(), (SDL_bool)resizable);
+#endif
 
     // Even in fullscreen mode, temporarily show the window in windowed mode to centre it.
     // This ensures that the window will be centred correctly when exiting fullscreen mode.
@@ -274,17 +274,6 @@ bool Gosu::Window::tick()
         switch (e.type) {
             case SDL_WINDOWEVENT: {
                 switch (e.window.event) {
-                    #ifdef GOSU_IS_MAC
-                        // Workaround for https://github.com/gosu/gosu/issues/458
-                        // "Resize" the window to its current dimensions after it is shown.
-                        // Otherwise it will be black on macOS 10.14 (Mojave) until the user moves it around.
-                        // TODO: Since this affects `brew install supertux` as well, maybe file an SDL bug?
-                        case SDL_WINDOWEVENT_SHOWN: {
-                            resize(this->width(), this->height(), fullscreen());
-                            break;
-                        }
-                    #endif
-
                     case SDL_WINDOWEVENT_SIZE_CHANGED: {
                         if (pimpl->resizable && (width() != e.window.data1 || height() != e.window.data2)) {
                             resize(e.window.data1, e.window.data2, fullscreen(), true);
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -162,24 +162,37 @@ void Gosu::Window::resize(unsigned width, unsigned height, bool fullscreen)
     if (fullscreen) {
         actual_width  = Gosu::screen_width(this);
         actual_height = Gosu::screen_height(this);
-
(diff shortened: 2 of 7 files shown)


jlnr:
Thanks for the PR. We can't make *all* windows resizable because existing games won't expect that the size and aspect ratio change while the game is running. I also think that *if* we add window resizing, the game should learn about it through a well-defined interface.

But the first question is -- what is the use case for this?

cyberarm:
Got this working* however it's a bit messy as I'm not to familiar with C++.

*Added an optional parameter to `shared_window(bool resizable = false)` and duplicated the Window initializer as I couldn't figure out how to do it cleanly... 😞 

Tested using this gist: https://gist.github.com/cyberarm/1209748671dc3a4c2df1975cabc45745

Kjarrigan:
I think its not that much code and could be usefull in some non-game applications. In theory we could add all kinds of logic but acutally it could work as it is now + one additional bool flag. The interface is already there as I can just overload resize. And wether the content should be scaled (or just displayed in the middle with a black border around it or only scale the game and not the ui or ...) is up to the developer.
It could look like this:

```ruby
class Gosu::Window
  def resizable?; false; end
end

class MyResizeableWindow < Gosu::Window
  def resizable?; true; end

  def resize(width, height, fullscreen=false)
    # do some fancy checks to like checking for a minimum size
    # calculate the scaling but keep the ratio and translate to the center of the screen.
    super
  end

  def draw
    Gosu.scale @scale_factor_from_resize_x, @scale_factor_from_resize_x do
      Gosu.translate @offset_from_resize_x, @offset_from_resize_y do
        # my usual content
      end
    end
  end
end
```

jlnr:
Hmm. Why does this PR have such a long diff? Can you please look into it, maybe try rebasing this branch?

I can also think of games where resizing a window makes sense, so 👍 for this functionality.

`resizable: true` can be a constructor flag that Gosu translates into a call to `SDL_SetWindowResizable` (`Window#initialize` already accepts an options hash). I don't think we need to support windows that are "sometimes resizable" so I'm against using a callback method for this.

@Kjarrigan I agree that resizable windows need a way to enforce minimum and maximum sizes *and* to maintain certain aspect ratios. Overriding `resize` sounds good for this :)

Kjarrigan:
Your right. Forgot that you can set the Window flags through seperate calls. So I guess @cyberarm please rebase against master, remove the printf and make add the flag to the window through an option.

Is there a way to test this on the CI? I read in the SDL doku, that there is a way to simulate SDL Events. Maybe we could come up with a simple interface or at least add an "interactive test". Also we need to add some hint in the docs about this.

jlnr:
Since most of the functionality is provided by SDL, I wouldn't invest too much in a unit test for this.

I think there are still some conceptual issues, though. If you run `Gosu::Window(320, 240, fullscreen: true).show`, then Gosu will scale up the window contents in software. (Nobody likes resolution switching.) Should Gosu still do this for resizable windows? What if the fullscreen resolution (=desktop resolution) is larger than the maximum size allowed by `MyGameWindow#resize`?

jlnr:
Also we need to check (manually) if an implementation of `resize` that restricts the window to a certain aspect ratio feels natural when resizing the window from all directions.

cyberarm:
@jlnr @Kjarrigan 
After fiddling with git for a while I found out I'd originally rebased against my fork instead of upstream, so a `rebase -i upstream master` and `git push origin master --force` later I think it is corrected... 🤞 

Now the actual changes are apparent.

So right now the implementation is:
``` ruby
Gosu::Window(width, height, fullscreen = false, update_interval = 16.6)
Gosu::Window(width, height, fullscreen = false, resizable = false, update_interval = 16.6)

# returns bool indicating window's resizability as set in constructor.
resizable

# Method that is called whenever the window size changes (excluding resize())
# Intended to prevent possible need to poll the window for its size
size_changed

# calls SDL_SetWindowResizable(shared_window(), width, height)
minimum_size(width, height)
```


Not sure how to make Ruby aware of these methods, is there documentation somewhere?

Gosu/Window.hpp
```diff
@@ -50,11 +50,21 @@ namespace Gosu
...
         Window(unsigned width, unsigned height, bool fullscreen = false,
             double update_interval = 16.666666);
+        Window(unsigned width, unsigned height, bool fullscreen = false, bool resizable = false,
+            double update_interval = 16.666666);
         virtual ~Window();
 
         unsigned width() const;
         unsigned height() const;
         bool fullscreen() const;
+        bool resizable() const;
```
Kjarrigan: Don't think this is necessary.

Gosu/Window.hpp
```diff
@@ -50,11 +50,21 @@ namespace Gosu
...
+
+        //! Called when the size of the window has changed.
+        virtual void size_changed() {};
```
Kjarrigan: Can be removed - the "callback" via an overloaded `` is good enough

Gosu/Window.hpp
```diff
@@ -50,11 +50,21 @@ namespace Gosu
...
+
+        //! Sets the minimum size the window can be.
+        void minimum_size(unsigned width, unsigned height);
```
Kjarrigan: Don't need this as well. This could be done in resize as well as checks for maximum size, ratio and adjustments to your drawing code to the new size.

Gosu/Window.hpp
```diff
@@ -50,11 +50,21 @@ namespace Gosu
         //! function.
         Window(unsigned width, unsigned height, bool fullscreen = false,
             double update_interval = 16.666666);
+        Window(unsigned width, unsigned height, bool fullscreen = false, bool resizable = false,
```
Kjarrigan: Not sure about this one - I personally would just add it behind the update_interval but Overloading in C++ is fine i guess. In Ruby on the other hand I think this would be a good opportunity to introduce an options-Hash. So I can imagine something like:
```ruby
class Gosu::Window
  def initialize(width, height, fullscreen: false, update_interval: 16.666666, resizable: false)
end
```

What do you think @jlnr (btw was there a reason you did define the options-Hash logic in ruby.i instead of patches?)

Kjarrigan:
> Not sure how to make Ruby aware of these methods, is there documentation somewhere?

It is - in the Wiki topic - Hacking Gosu: https://github.com/gosu/gosu/wiki/Hacking-Gosu

jlnr:
I've added this to the 0.15 project because 0.14 is already way behind schedule (wanted to finish this in 2017 :) ).

Kjarrigan:
I think thats ok - 0.15 doesn't mean it gets released in 2030 or something just not in the next planned release where we're already polishing :smile: 

@cyberarm - I saw that you already cleaned up your code. Did you look in the SWIG part I linked above? Or do you need more help to finish you PR?

cyberarm:
@jlnr @Kjarrigan, I ran the rake task however it errored on one of the patches:
``` text
swig -c++ -ruby -autorename -o src/RubyGosu.cxx ext/gosu/gosu.i
patch --no-backup-if-mismatch -p0 <ext/gosu/gosu_SWIG_RENAME_PATCH.patch
patching file src/RubyGosu.cxx
Hunk #1 FAILED at 2912.
Hunk #2 succeeded at 3220 with fuzz 2 (offset 293 lines).
Hunk #3 succeeded at 3235 with fuzz 2 (offset 293 lines).
Hunk #4 succeeded at 3249 with fuzz 2 (offset 293 lines).
Hunk #5 succeeded at 3267 with fuzz 2 (offset 290 lines).
Hunk #6 succeeded at 3278 with fuzz 2 (offset 287 lines).
Hunk #7 succeeded at 11715 with fuzz 2 (offset 200 lines).
1 out of 7 hunks FAILED -- saving rejects to file src/RubyGosu.cxx.rej
rake aborted!
Command failed with status (1): [patch --no-backup-if-mismatch -p0 <ext/gos...]
```

P.S. I forgot to fetch upstream before pushing to my fork, so I had to force push again so that's why the commits are all clumped together.

jlnr:
I've updated the line offsets in the diff files since I'm not sure if `patch` is as tolerant on Linux as it is on macOS (GNU vs BSD, who knows); I've also run `rake swig` again with SWIG 3.0.12. I'm not sure which one caused `rake swig` to fail for you, but now it should be up-to-date :)

Kjarrigan:
Well, Worked for me on Linux so I guess its more of an Windows issue ;-)

cyberarm:
Resizable argument is now accessible in Ruby version (Quickly tested using `Gosu::Window.new(100,100, :resizable => true).show`)

I still can't get patch working on Linux Mint or Windows (with GnuWin32 patch), but at a glance looks like I don't need to.. Thanks @jlnr for fixing it.

cyberarm:
Okay, got `rake swig` working 😃 Apparently `patch` ships with Git for Windows but its not in path by default... 

Tested this time with a slightly more complicated project: https://github.com/cyberarm/visual_plotter

cyberarm:
@jlnr @Kjarrigan
I'm starting over on this feature from upstream 55e9c2b2dff0b415e3119831062b5516674967f2  with work being done in [this branch](https://github.com/cyberarm/gosu/tree/make_window_resizable).

Should I just switch this pull request's target branch or open another pull request?

jlnr:
@cyberarm The *target* branch is fine (master). I'm not sure if you can change the *source* branch. However, you can just force-push your commits from the new branch to cyberarm/gosu/master, which will update this PR.

src/Window.cpp
```diff
@@ -85,34 +85,37 @@ struct Gosu::Window::Impl
...
 
-Gosu::Window::Window(unsigned width, unsigned height, bool fullscreen, double update_interval)
+Gosu::Window::Window(unsigned width, unsigned height, bool fullscreen, double update_interval, bool resizable)
 : pimpl(new Impl)
 {
     // Even in fullscreen mode, temporarily show the window in windowed mode to centre it.
     // This ensures that the window will be centred correctly when exiting fullscreen mode.
     // Fixes https://github.com/gosu/gosu/issues/369
     // (This will implicitly create graphics() and input(), and make the OpenGL context current.)
+    SDL_SetWindowResizable(shared_window(), (SDL_bool)resizable);
```
cyberarm: Travis's SDL must be lower than SDL 2.0.5... Build fails here.

jlnr: There's an issue for upgrading Gosu's CI environment #478 — but Ubuntu 16.04 still ships with 2.0.4, so that won't help. To make the CI happy, this should be wrapped in an `#if` block like other "recent" SDL features.

dfl:
I'm interested in this feature... can also test on Mac if needed.

cyberarm:
@jlnr @Kjarrigan I believe this is technically complete, window can be resized without glitches (on Manjaro Linux.) I left in the `Window.resizable` in case, e.g. someone needs to know if the window is resizable and they're using a library on top of gosu.

Should I add `Window.resize` to swig so ruby can override it?

cyberarm:
@dfl This should be ready for testing, it requires SDL 2.0.5 to work.
For ruby:
* Clone my fork: https://github.com/cyberarm/gosu
* Enter directory and run:
* `GOSU_RELEASE_VERSION="9.9.9" rake gem`
* `gem install pkg/gosu-9.9.9.gem`

jlnr:
I'll try it once I find the time — I am mostly concerned about https://github.com/gosu/gosu/pull/437#issuecomment-372815012, how does this interact with Gosu's automatic scaling?

cyberarm:
@jlnr As implemented right now, if the window is resizable and the user resizes it with the window manager, scaling will be ignored.
It would require some reworking, but it could be made to optionally support keeping the aspect ratio.

I was acting under the assumption that if the window is resizable then the developer will handle repositioning and scaling elements how they want.

jlnr:
@cyberarm I think the idea would not be to add another flag for keeping the aspect ratio, but instead letting the user implement this kind of code themselves, by making `resize` an overridable callback as here - https://github.com/gosu/gosu/pull/437#issuecomment-372716665

This also allows for other things such as ensuring the window size is always a multiple of 50 for certain tile-based UIs, etc. I know that the Windows API has a message where windows are both informed about their new size, but they can also modify it at the same time: https://docs.microsoft.com/de-de/windows/desktop/winmsg/wm-windowposchanging - not sure if/how that's possible with SDL to be honest :)

jlnr:
Works great in windowed mode on macOS! I still have to look at the `resize` callback, but one unexpected issue is that once the window goes fullscreen, it cannot be un-fullscreened anymore, the button is disabled.

jlnr:
I've changed a few minor things to be more consistent with the rest of Gosu. The only *real* change is that I've removed the `resizing` parameter from `Window::resize`. Is this necessary on Linux? It works fine without it on macOS, will try Windows later. It's probably a little wasteful to resize the window from within the resize callback, but it's a bit awkward to have an internal parameter as part of Window's public API.

I've tried to build a `resize` callback that makes sense, but it's super fiddly because the `resize` name is already taken by the C++ API, the SDL doesn't send any events for live resizing, etc. I think at this point it's better to have no specific API for this than a bad one. So basically, this PR is ready for merging from my side as soon as Windows and Linux have been (re-)tested with my changes.

jlnr:
Welp okay, resizing from within the resize callback looks really broken on elementaryOS. I'll send another commit to undo my damage.

pmer:
Sorry to chime in—just wanted to say that a lot of games on Windows and macOS have resizable windows. You see this in both AAA games—e.g., Blizzard productions—and smaller, single-developer games—e.g., Silicon Zeros. Maybe you encounter it in situations where the developer has been able to put in extra time for polish?

cyberarm:
@jlnr I believe I fixed this by added a `resizing` bool to the `Gosu::Window::Impl` struct, instead of having an additional argument for `resize`, which is what it was.

If you want to double check this on Mac to make sure it works as expected?

jlnr:
@cyberarm Ah thanks, much better. The remaining issue that I saw on my Mac with this approach is that scaling in fullscreen mode is broken now. I wonder if the workaround for the (now fixed) SDL bug did actually matter? Need to look at this when I have a free day :)

jlnr:
I think all the combinations of fullscreen and resizable look good now, even when manually calling `Window::resize` from the game's code.

jlnr:
Merged. Thanks again for the contribution, and @Kjarrigan for the review!

Issue #390, opened 2017-03-26
Gosu::Song is broken without Gosu::Windo


jlnr:
It is not intuitive that Gosu::Song relies on a window being shown, since it does not require a window in its constructor anymore:

http://stackoverflow.com/questions/42565727/gosusong-in-ruby-only-playing-first-note

Easy fix: Expose and document the `Song#update` method that already exists in C++ - it is what Gosu::Window uses.

Issue #117, opened 2012-03-12
Merge Song and Sample into a single class


jlnr:
See discussion here https://github.com/jlnr/gosu/issues/116

How to name it? Sound? SampleInstance could be renamed according to http://www.libgosu.org/cgi-bin/mwf/topic_show.pl?tid=624

Kjarrigan:
I'll look into this. This is be one of the oldest issues but resolving this might resolve other issues (#390) as well or at least make things a little easier here and there ( #147 , #264 ).

jlnr:
Hmm. I don't actually know how this issue would interact with #390. It's bad enough that `Gosu::Window` has to keep `Song.current_song` running, but how would it work if there could be any number of streaming Sounds at the same time? (Basically, Sample would become `Sound.new(filename)` and Song would be `Sound.new(filename, streaming: true)`)

jlnr:
Also, there is actually one supported platform where `Gosu::Song` is magic: iOS. It uses a completely different API for the current song because that way it's hardware-accelerated. But if the merger works well on "normal" platforms, I'm sure I can find a way to make it work on iOS as well :)

jlnr:
A quick braindump of suggestions:
- Sample is renamed into Sound and gains a `streaming:` constructor parameter. It then dynamically uses one of severeal implementations: the old Sample code, and whatever code Song uses when streaming a file. (Song already has a tiny class hierarchy for its implementations, with bizarro class names for historical reasons.)
- Song (for backward compatibility) is a simple wrapper around a Sound and a SampleInstance (should be renamed to Channel while we are at it - #314). When you call `Song#play`, it stores the Channel so that it can implement `Song#stop` etc.
- All *streaming* sounds add themselves to a global array while they are playing, and Song::update() (C++ only) moves to Sound::update() and lets *all* of these Sounds stream more data. Gosu::Window doesn't even need to be changed.

The alternative is to work with threads to fix #390, but I think that would too much new code in addition to all the refactoring action, and should be a separate branch.

Kjarrigan:
> Song (for backward compatibility) is a simple wrapper around a Sound and a SampleInstance (should be renamed to Channel while we are at it - #314). When you call Song#play, it stores the Channel so that it can implement Song#stop etc.

Should this wrapper be available in C++ and Ruby?

> The alternative is to work with threads

I definitely won't work with threads.

[...]

PR #441, opened 2018-03-13
[WIP] Audio 2.0

Kjarrigan:
As the Sample-Song-Merge is quite a big thing (at least in my opinion) I already start this PR even if there is not much to show at the moment BUT I think its already worth discussing.

So here we are. I basically only changed Audio.hpp and test_audio.rb so far to show how I think the final interface would/should look like after merge.

### DIFF
--- a/Gosu/Audio.hpp
+++ b/Gosu/Audio.hpp
@@ -11,10 +11,15 @@
 
 namespace Gosu
 {
-    //! Sample::play returns a Channel that represents the sound currently being played.
+    //! Sound::play returns a Channel that represents the sound currently being played.
     //! This object can be used to stop sounds dynamically, or to check whether they have finished.
     class Channel
     {
+        class BaseData;
+        class ModuleData;
+        class StreamData;
+        std::unique_ptr<BaseData> data;
+
         mutable int channel, token;
 
     public:
@@ -33,33 +38,38 @@ namespace Gosu
         //! Calling this twice, or too late, does not do any harm.
         void stop();
 
+        //! Returns the current volume of the song.
+        double volume() const;
         //! \param volume Can be anything from 0.0 (silence) to 1.0 (full volume).
         void set_volume(double volume);
+
         //! \param pan Can be anything from -1.0 (left) to 1.0 (right).
         void set_pan(double pan);
         //! \param speed Use 1.0 for normal playback speed.
         void set_speed(double speed);
     };
-    
-    //! A sample is a short sound that is completely loaded in memory, can be
-    //! played multiple times at once and offers very flexible playback
-    //! parameters. Use samples for everything that's not music.
-    class Sample
-    {
-        struct SampleData;
-        std::shared_ptr<SampleData> data;
 
+
+    //! A sound handles all your audio needs from short samples (e.g. explosions) to long songs (e.g. background music)
+    //!
+    //! Use streaming = false for everything that's not music. (Maybe add a more detailed explanation why this is better,
+    //! because some the disadvantages of song might dissappear with the merge?)
+    class Sound
+    {
+        struct SoundData;
+        std::shared_ptr<SoundData> data;
+        std::vector<Channel> current_songs;
     public:
         //! Constructs an empty sample that acts as if the song had a length of 0.
-        Sample();
+        Sound();
         
         //! Constructs a sample that can be played on the specified audio
         //! system and loads the sample from a file.
-        explicit Sample(const std::string& filename);
+        explicit Sound(const std::string& filename, bool streaming = false);
         
         //! Constructs a sample that can be played on the specified audio
         //! system and loads the sample data from a stream.
-        explicit Sample(Reader reader);
+        explicit Sound(Reader reader, bool streaming = false);
         
         //! Plays the sample without panning.
         //! \param volume Can be anything from 0.0 (silence) to 1.0 (full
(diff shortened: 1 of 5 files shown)


Gosu/Audio.hpp
```diff
@@ -33,33 +38,38 @@ namespace Gosu
...
+
+    //! A sound handles all your audio needs from short samples (e.g. explosions) to long songs (e.g. background music)
+    //!
+    //! Use streaming = false for everything that's not music. (Maybe add a more detailed explanation why this is better,
+    //! because some the disadvantages of song might dissappear with the merge?)
+    class Sound
+    {
+        struct SoundData;
+        std::shared_ptr<SoundData> data;
+        std::vector<Channel> current_songs;
```
Kjarrigan: Not sure about the naming but thats basically the global variable you mentioned in #117

Gosu/Audio.hpp
```diff
@@ -33,33 +38,38 @@ namespace Gosu
...
     public:
         //! Constructs an empty sample that acts as if the song had a length of 0.
-        Sample();
+        Sound();
         
         //! Constructs a sample that can be played on the specified audio
         //! system and loads the sample from a file.
-        explicit Sample(const std::string& filename);
+        explicit Sound(const std::string& filename, bool streaming = false);
```
Kjarrigan: Are there named parameters in c++. Googling resulted in different opionions on how to and if at all...

Gosu/Audio.hpp
```diff
@@ -80,59 +90,7 @@ namespace Gosu
         //! or low values. Use 1.0 for normal playback speed.
         Channel play_pan(double pan, double volume = 1, double speed = 1,
             bool looping = false) const;
-    };
-
-    //! Songs are less flexible than samples. Only Song can be played at any given time,
-    //! and there is no way to control its pan (stereo position) or speed.
-    class Song
```
Kjarrigan: Should Song be available in C++ for backwardscompat as well?

Gosu/Audio.hpp
```diff
@@ -80,59 +90,7 @@ namespace Gosu
...
-        //! Returns true if the song is currently playing.
-        bool playing() const;
-        //! Returns the current volume of the song.
-        double volume() const;
-        //! \param volume Can be anything from 0.0 (silence) to 1.0 (full
-        //! volume).
-        void set_volume(double volume);
-        
         //! Called every tick by Window for management purposes.
         static void update();
```
Kjarrigan: Is there a good reason why this can't be available in Ruby as well? It would at least to some degree solve issue #390 and has no negative side effects as far as I can tell. This way we can get rid of the window in the tests and the guy who used gosu just to play music needs some kind of mainloop one way or the other. I'm not sure if playing sounds without window should work out of the box in Gosu, as its a "game development" lib :smile:

lib/gosu/compat.rb
```diff
@@ -103,24 +103,26 @@ def self.from_text(*args)
   end
 end
 
-# No need to pass a Window to Sample.
-class Gosu::Sample
-  alias initialize_without_window initialize
-
+# No need to pass a Window to Sample and even better don't use Sample at all.
+class Gosu::Sample < Gosu::Sound
```
Kjarrigan: Inheritance because we got the double deprecation, but maybe just replace the whole construct with deprecate_const?

lib/gosu/compat.rb
```diff
@@ -103,24 +103,26 @@ def self.from_text(*args)
...
       Gosu.deprecation_message("Passing a Window to Sample#initialize has been deprecated in Gosu 0.7.17.")
     end
-    initialize_without_window(*args)
+    super(*args, strea)
   end
 end
 
-# No need to pass a Window to Song.
+# No need to pass a Window to Song and even better don't use Song at all.
 class Gosu::Song
```
Kjarrigan: Depending on whether or not the c++ counterpart of Song should still exists this might need inherit from Gosu::Sound too.

Gosu/Audio.hpp
jlnr: Not really. You can build them yourself but I try to use only basic features in Gosu (both C++ and Ruby).

lib/gosu/compat.rb
```diff
@@ -165,7 +167,6 @@ module Gosu
   module Button; end
   
   # Channel was called SampleInstance before Gosu 0.13.0.
-  SampleInstance = Channel
```
Kjarrigan: Thats not necessary

Gosu/Audio.hpp
jlnr: Hmmm...I would say yes. At least for now :)

test/test_audio.rb
```diff
@@ -10,8 +10,8 @@ class TestAudio < Minitest::Test
   def test_sample
     skip_on_appveyor
 
-    sound = Gosu::Sample.new(media_path("0614.ogg"))
-    channel = sound.play(1, 1, true)
+    sound = Gosu::Sound.new(media_path("0614.ogg"))
+    channel = sound.play(volume: 1, speed: 1, looping: true)
```
Kjarrigan: I thought, that if we already "drastically" change the naming part we might as well make this a hashed list as well?

Gosu/Audio.hpp
jlnr: Good point. Maybe there's some way in which Gosu could spawn a thread that keeps all sounds running, but in the meantime, Ruby users should have access to this.

test/test_audio.rb
```diff
@@ -50,68 +50,54 @@ def test_sample
...
-    win = SongTestWindow.new
-    
-    assert_nil Gosu::Song.current_song
+    # assert_nil Gosu::Sound.current_songs
     
-    song = Gosu::Song.new(media_path("0830.ogg"))
+    song = Gosu::Sound.new(media_path("0830.ogg"), streaming: true)
 
-    song.play(true)
+    channel = song.play(looping: true)
```
Kjarrigan: Really not sure of this, but most folks already adapted the Sample.play = Channel logic, so why not make it the same for "Songs"

lib/gosu/compat.rb
jlnr: +1 deprecate_const, Gosu::Sound's interface should be compatible with Sample's

lib/gosu/compat.rb
jlnr: [Looks like that's right](https://searchcode.com/?q=SampleInstance&lan=32), I was worried about people writing `if x.is_a? Gosu::SampleInstance`.

test/test_audio.rb
jlnr: I know that Ruby 2.1+ supports some form of named arguments on the language level, and I've been meaning to look into this. I think your proposed interface should be the long-term goal, but I would still provide the old interface for compatibility (so that we can just `Sample = Sound` for now; also makes the C++ bridging a bit easier).

jlnr: Also, this would be a good way to get rid of `play_pan` altogether. I think the issue is that `play` used to be louder than `play_pan(0)` at some point in time; I don't think that's true anymore, and even if it is, then the default value for `pan` could just be `nil` instead of `0` to distinguish both cases.

Gosu/Audio.hpp
jlnr: `streaming_sounds`? This doesn't have to be part of `Audio.hpp`, it can be a static variable in `Audio.cpp`.

test/test_audio.rb
Kjarrigan: I thought about doing it the same way you handled Image.from_text. There you already deprecated a argument list and hinted towards the option-hash. if we internally use a Hash (to be backward compat to Ruby <2.0) or named parameters is just a matter of personal preference (and maybe a bit about more or less code). Ruby 2.1 only introduced _require-named-parameters_ which also reduces a bit of code by doing some checks and raising an error automatically.
So I guess you're the one who defines whats the lowest Ruby Version Gosu should support?

lib/gosu/compat.rb
Kjarrigan: This still works and triggers the deprecation warning too.

test/test_audio.rb
jlnr: I was planning to raise the minimum version to 2.3 right after Gosu 1.0. Some Gosu-related gems are still only compatible with Ruby 1.9.3(!) as far as I know. But I don't think we really need *required* arguments in `Sound#play` so a simple hash sounds fine as in `Image.from_text` sounds good enough :)

jlnr: Another random thought. We should test whether `play_pan(0)` and `play` sound the same *with a stereo sample*, not just good old Beep.wav :)

jlnr:
And one more thought. Right now Gosu reserves one dedicated OpenAL channel for songs. That prevents a situation where `Song.play` silently fails when many samples are playing already.

One clean solution would be to pass an optional priority into every `play` call and samples with a higher priority can kick samples with a lower priority out even while they're playing. But...I guess we can push this issue back until someone actually runs into this situation :)

jlnr:
Another thing that I'll have to rethink with this PR is the "automagic handling" of Gosu::Song in the iOS/UIKit port:

https://github.com/gosu/gosu/blob/master/src/GosuViewController.cpp#L108-L128

Both the code and logic are ancient. I need to work on that part anyway because it's using deprecated APIs.

Issue #413, opened 2017-09-04
Subtractive drawing mode


KoBeWi:
Right now, Gosu supports _:additive_ and _:multiply_ drawing modes. I've been using them extensively recently and I realized something like _:subtractive_ (or _:subtract_) mode, which subtracts color values, would be useful too.

jlnr:
Absolutely. I think it would be even cooler to have the full suite of Photoshop blend modes available (this is from the PDF file format reference, but I think these are the same as in Photoshop):

https://www.adobe.com/content/dam/Adobe/en/devnet/pdf/pdfs/pdf_reference_archives/blend_modes.pdf

I'm not sure how many of them can be implemented with just plain OpenGL; maybe some of them need shaders.

Issue #198, opened 2014-01-22
Feature Request: Create Sample from Array or binary String


runnerpack:
I would like to do some synthesizer experimentation in my favorite language (Ruby) with my favorite game library (you can probably guess... :wink: )

One can already make an Image from a pixel "blob", so why not Samples (and Songs?) from a PCM "blob"?

Not really a priority, but you may as well add it to the ol' ToDo list and work on it when you're bored, right? :wink:

Thanks, jlnr!

runnerpack:
Not that you have to explain yourself, since it's your project, but I'm curious why you removed this from the 1.0 todo list…  (or maybe you did explain your reasoning somewhere and I didn't catch it…)
When I first proposed it, you said it would be easy, and that you were in favor of it *shrug*

jlnr:
I'll explain myself anyway: Gosu 1.0 is mostly about having an API that is "good enough" so that there can be a period of long stability (in retrospect, Gosu 0.7.x should have been 1.0.x, given how long-lived it was). My mental roadmap looked something like this:

1. one 0.x release that fixes some limitations of Gosu::Input (#524 is shaping up to be just that).
2. one 0.y releases that closes all the audio issues (including this one), mostly because #441 is a big change to the audio interface and it seemed reasonable to do it all at once.
3. maybe named arguments in Image#draw etc.

After I ran out of time to push _any_ of these areas forward, I've decided to bump up the version number to 1.0 right after merging #524. So this is not really about de-prioritizing this particular feature (which has been stuck for six years anyway >_<). It's more about admitting that the pace of development is incredibly slow, and that the current version is more stable than its version number indicates.

runnerpack:
That all makes perfect sense.  Thanks for the information!  See you in 2026 ;)

PR #443, opened 2018-03-22
Add Image.draw(options) and tests (and remove draw_rot)

Kjarrigan:
Hi @jlnr ,

sorry for another PR but while looking through ashton and working on my little gosu game I stumbled across some things with the images.... Especially the long argument list bothered me quite a while and now that I know the internals a bit better I just changed it :smile:
Here one general question remains for me - was there a specific reason you coded the other options (e.g. from_text) in the SWIG-File instead of Ruby? The latter is much more compact and keeping the SWIG stuff smaller is a good idea I think. When we drop Ruby < 2.x support we can make this even shorter with the builtin named parameters (which do all kinds of checks like raising if you have a typo in your param name).

Some of the changes in test_helper are stolen from #433 so I'll rebase if either one is merge first (did you have a change looking at the open questions over there?)

The _replace_-blend mode was something which looked important for ashton and wasn't hard to add (and more blend modes are already requested #413) so I thought I just add it.

Finally I added tests for all the things (at least I hope I didn't miss anything - I just looked through libgosu.org/rdoc from top to bottom). To keep the repo small I tried to reuse as many images as possible but maybe I should look for a better image for the retro/tileable-scaling tests.


src/RenderState.hpp
```diff
@@ -37,14 +37,17 @@ struct Gosu::RenderState
         }
     }
     
-    void apply_alpha_mode() const
+    void apply_blend_mode() const
```
Kjarrigan: alpha_blend_mode confused me a bit the first time I looked through the files because you actually change not only the Alpha-Channel, so I renamed it.

test/test_image.rb
```diff
@@ -0,0 +1,205 @@
...
+    win = ScreenshotWindow.draw(100, 100) do |w, h|
+      c = 0xff_ffffff
+      @image ||= Gosu::Image.new(media_path('triangle-50.png'))
+      @image.draw_as_quad(0, 0, c, w, 0, c, w, h, c, w-35, 15, c, 0, :default)
+
+      # BUG: Drawing the quad from bottom to top is buggy (last Vertex is always wrong)
+      # should     4--3
+      # be:       /    \
+      #          1------2
+      @image.draw_as_quad(0, w, c, w, h, c, w-15, h/2, c, 15, h/2, c, 0, :default)
```
Kjarrigan: Not sure about this one - But the one vertex is always off? I tried: 
1-2  | 3-4 | 1-2
4-3  | 1-2 | 3-4

Kjarrigan: Update - might be an Linux thingy. Because thats the only green system on CI - every other platform failed this test (with my wrong image)....

Kjarrigan:
I added `Image.from_blob` - mainly for viewing the blob differences from the CI but I can imagine some cool images/textures could be generated from random strings as well. ( In the spirit of #198 ).

Also a small bugfix discovered by simplecov (the gl_tex_info test was never called)...

Gosu/GraphicsBase.hpp
```diff
@@ -26,7 +26,9 @@ namespace Gosu
         //! to the old color's channels.
         AM_ADD,
         //! The color's channels will be multiplied with each other.
-        AM_MULTIPLY
+        AM_MULTIPLY,
+        //! The color will completly replace the current one.
```
jlnr: *completely

What is this mode good for? I've thought about extending the AlphaMode enum to cover all of [Adobe's blend modes](https://helpx.adobe.com/photoshop/using/blending-modes.html) because I'm familiar with them from Photoshop, but I'm not sure what `AM_REPLACE` could be used for.

ext/gosu/gosu.i
```diff
@@ -658,6 +661,15 @@ namespace Gosu
         std::unique_ptr<Gosu::ImageData> image_data = $self->data().subimage(x, y, w, h);
         return image_data.get() ? new Gosu::Image(std::move(image_data)) : nullptr;
     }
+
+    %newobject from_blob;
+    static Gosu::Image* from_blob(const std::string& blob, int width, int height)
+    {
+        Gosu::Bitmap bitmap(width, height);
+        memcpy(bitmap.data(), &blob[0], width * height * 4);
```
jlnr: Since we know how long `blob` is, we can perform a range check here instead of reading from possibly invalid memory.

lib/gosu/compat.rb
```diff
@@ -101,6 +101,38 @@ def self.from_text(*args)
...
+  alias _draw draw
+  def draw(x, y, z, *args)
+    if args.last.is_a?(Hash)
+      opts = args.pop
+      scale_x = args[0] || opts.fetch(:scale_x, 1)
+      scale_y = args[1] || opts.fetch(:scale_y, 1)
+      color   = args[2] || opts.fetch(:color, 0xff_ffffff)
+      mode    = args[3] || opts.fetch(:mode, :default)
+
+      # Make draw_rot obsolete as the params are bascially the same besides the rotation-specific values
```
jlnr: *basically

rdoc/gosu.rb
```diff
@@ -390,6 +390,18 @@ def self.load_tiles(source, tile_width, tile_height, options = {}); end
     # @return [Image?] an image that represents a portion of the containing image
     def subimage(left, top, width, height); end
 
+    ##
+    # Returns an image from a binary string of packed RGBA values. (e.g. from (Image#to_blob))
+    #
+    # @param blob [String] a string of any length (preferably width * height * 4)
+    # @param width [Ingeger] the width of the resulting image
```
jlnr: *Integer

rdoc/gosu.rb
```diff
@@ -390,6 +390,18 @@ def self.load_tiles(source, tile_width, tile_height, options = {}); end
     # @return [Image?] an image that represents a portion of the containing image
     def subimage(left, top, width, height); end
 
+    ##
+    # Returns an image from a binary string of packed RGBA values. (e.g. from (Image#to_blob))
+    #
+    # @param blob [String] a string of any length (preferably width * height * 4)
```
jlnr: Not just preferably! :)

jlnr: It should mention that this should be a binary string, plus the byte order

src/RenderState.hpp
jlnr: Good point. I guess the whole enum should be named `Gosu::BlendMode` with `BM_ADD` etc. as its members... :/

Gosu/GraphicsBase.hpp
jlnr: (I can think of use cases when drawing into an RGBA buffer; `Gosu::Bitmap::insert` is basically `AM_REPLACE`. But when you are drawing to the screen, you can't replace the destination pixels with translucent pixels.)

test/test_helper.rb
```diff
@@ -35,6 +36,19 @@ def similar?(img, threshold=0.90)
 end
 
 module TestHelper
+  class ScreenshotWindow < Gosu::Window
+    include TestHelper
+
+    def self.draw(w, h)
+      win = new(w,h)
```
jlnr: +space between operators

test/test_helper.rb
```diff
@@ -49,20 +63,48 @@ def skip_on_travis
   end
 
   def skip_on_ci
-    skip if ENV["APPVEYOR"] or ENV["TRAVIS"]
+    skip if runs_on_ci?
+  end
+
+  def runs_on_ci?
```
jlnr: Ruby doesn't usually have an `s` at the end of verbs (e.g. `equal?`). `run_on_ci?`, `on_ci?`, simply `ci?`? Not sure

test/test_image.rb
jlnr: Shouldn't the first parameters be `0, h, c`?

jlnr:
Re `Image.draw(x, y, z, options)`: I like it! The long argument list has been annoying for ten years(!) now, see this discussion thread: https://www.libgosu.org/cgi-bin/mwf/topic_show.pl?pid=14

tl;dr my plan was to build `image.scale(5).rotate(5, 0.5, 0.5).draw(x, y, z)` instead of using an arguments hash. The idea here was to make it clear in which order operations are applied (scale first, rotation afterwards), and also to implement it without any extra object allocations, because Ruby GC pauses used to be a big problem with Gosu.

However, Ruby supports named arguments now, and I don't think it's a good idea to fight the language design. So let's do it!

As for the details: I'm not sure if `:center_x` and `:center_y` shouldn't work even without `:angle`; maybe the presence of `:angle` should only change their default values from 0.0 to 0.5?

What do you think about having `scale:` as a shorthand for `scale_x:` and `scale_y:`? I guess a shortcut for `center:` should also be added for consistency then, even if it seems less useful.

jlnr:
Re: `Image.from_blob`, are you aware of how `Image.new(...)` already supports RMagick images as its input parameter? I've added it for consistency with the `Image(Bitmap)` constructors in the C++ API. But I think `from_blob` is actually much better, because it's much more discoverable, and RMagick has never been a popular companion library for Gosu. So :+1: for an explicit `from_blob` method, but it should use the same code path as the RMagick-based constructor, including the support for float-based RGBA binary strings (I think this might have been added for the TexPlay gem?).

jlnr:
Re: AM_REPLACE, I'm not convinced :) But I do agree with every bit of renaming from AlphaMode to BlendMode.

Kjarrigan:
Ok so many response....Hope I don't miss anything :smile: 

> tl;dr my plan was to build image.scale(5).rotate(5, 0.5, 0.5).draw(x, y, z) instead of using an arguments hash.

Didn't knew that. Definitely a considerable approach - a bit like the ActiveRecord-Relation chaining. This might be possible to add besides the parameter Hash which internally just converts it for the final draw.

> As for the details: I'm not sure if :center_x and :center_y shouldn't work even without :angle; maybe the presence of :angle should only change their default values from 0.0 to 0.5?

You're right. Drawing an image with the given (x,y) as the center rather than the upper-left-corner should be a thing and then the _center_-shortcut makes sense. We could even add a bit of abstraction by providing keywords instead of float numers like center: :middle (or :upperleft, :lowerright, etc.).

> What do you think about having scale: as a shorthand for scale_x: and scale_y:?

Absolutely.

> All kinds of typos....

I'll fix these. I guess I should look to some spell-checking addon for my editor :smile: 

> AM_REPLACE, I'm not convinced

I'm not totally sure how this exactly works as well but Ashton used this quite a lot when combining shaders (and had a hint that it has to be done "manually" because Gosu doesn't support it out-of-the-box). But I guess I just remove it for now until I figure out how all the bits are connected and then re-add it later if needed.

> AlphaMode to BlendMode. + AM_ to BM_

You're right - I'll change them too.

> Re: Image.from_blob, are you aware of how Image.new(...) already supports RMagick images in place of the filename parameter?

Yes and no. Thats what I tried first (in irb while trying to convert by AppVeyor dump to a real image) because I thought the conversion from RMagick to Gosu would use some kind of blob. But something like:

```ruby
Foo = Struct.new :to_blob, :columns, :rows
Image.new(Foo.new(Filebinread('dump'), 100, 100)).save('realimage.png')
```

didn't work :smile: 

> but it should use the same code path as the RMagick-based constructor, including the support for float-based RGBA binary strings (I think this might have been added for the TexPlay gem?).

I'll look into this.

ext/gosu/gosu.i
```diff
@@ -658,6 +658,32 @@ namespace Gosu
         std::unique_ptr<Gosu::ImageData> image_data = $self->data().subimage(x, y, w, h);
         return image_data.get() ? new Gosu::Image(std::move(image_data)) : nullptr;
     }
+
+    %newobject from_blob;
+    static Gosu::Image* from_blob(const std::string& blob, int width, int height)
```
jlnr: If `Gosu::Image` works (because it's default constructible now?), we should always just return images directly instead of working with pointers and `%newobject`.

lib/gosu/compat.rb
```diff
@@ -101,6 +101,70 @@ def self.from_text(*args)
       from_text_without_window(*args)
     end
   end
+
+  alias _draw_rot draw_rot
+  def draw_rot(*args)
+    Gosu.deprecation_message("Image#draw_rot is deprecated in Gosu x.y, use draw with an options hash instead, see https://www.libgosu.org/rdoc/Gosu/Image.html.")
```
jlnr: x.y = 0.14 :)

lib/gosu/compat.rb
```diff
@@ -101,6 +101,70 @@ def self.from_text(*args)
...
+      scale_x  = args[0] || opts.delete(:scale_x) || 1
+      scale_y  = args[1] || opts.delete(:scale_y) || 1
+      color    = args[2] || opts.delete(:color) || 0xff_ffffff
+      mode     = args[3] || opts.delete(:mode) || :default
+      center_x = opts.delete(:center_x)
+      center_y = opts.delete(:center_y)
+
+      # Shorthand for scale_x: scale_y:
+      scale = opts.delete(:scale)
+      scale_x, scale_y = scale, scale if scale
```
jlnr: `scale_x` should override `scale`, not the other way around

lib/gosu/compat.rb
```diff
@@ -101,6 +101,70 @@ def self.from_text(*args)
...
+
+      # Shortcut with symbols
+      center = opts.delete(:center)
+      if center
+        center_x, center_y = if center.is_a?(Symbol)
+          if DRAW_ALIGNMENTS.has_key?(center)
+            DRAW_ALIGNMENTS[center]
+          else
+            raise ArgumentError, "Unknown drawing alignmengt: #{center} should be a float or one of: #{DRAW_ALIGNMENTS.keys.join(', ')}"
```
jlnr: *alignment

lib/gosu/compat.rb
```diff
@@ -101,6 +101,70 @@ def self.from_text(*args)
...
+        else
+          [center, center]
+        end
+      end
+
+      # Make draw_rot obsolete as the params are basically the same besides the rotation-specific values
+      angle = opts.delete(:angle)
+
+      raise ArgumentError, "Unknown options: #{opts.keys} should be either of [angle, center, center_x, center_y, color, mode, scale, scale_x, scale_y]" unless opts.empty?
+      if angle or center_x or center_y
```
jlnr: I don't think it's intuitive that passing `center_x: 0.5` also sets `center_y: 0.5`, but leaving out `center_x` uses `center_y: 0`

jlnr: I still think changing the default value of `center_x`/`center_y` based on the presence of `angle` is more intuitive, if only slightly :)

lib/gosu/compat.rb
```diff
@@ -101,6 +101,70 @@ def self.from_text(*args)
...
+    :up => [0, -1],
+    :upper_right => [1, -1],
+    :left => [-1, 0],
+    :middle => [0, 0],
+    :right => [1, 0.0],
+    :bottom_left => [-1, 1],
+    :bottom => [0, 1],
+    :bottom_right => [1, 1],
+  }
+  def draw(x, y, z, *args)
```
jlnr: I wonder if executing relatively much Ruby code in every `Image#draw` call slows down existing Gosu games. I guess we should release a 0.14.0-pre1 gem this time and rewrite `Image#draw` in C++ if this is really an issue

rdoc/gosu.rb
```diff
@@ -390,6 +390,18 @@ def self.load_tiles(source, tile_width, tile_height, options = {}); end
...
+    # Returns an image from a binary string of packed RGBA values. (e.g. from (Image#to_blob))
+    #
+    # @param blob [String] a binary string with 1 byte per channel and pixel (RGBA) (width * height * 4)
+    # @param width [Integer] the width of the resulting image
+    # @param height [Integer] the height of the resulting image
+    #
+    # @return [Gosu::Image]
+    #
+    # @see Image#to_blob
+    def self.from_text(blob, width, height); end
```
jlnr: *from_blob :)

rdoc/gosu.rb
```diff
@@ -390,6 +390,18 @@ def self.load_tiles(source, tile_width, tile_height, options = {}); end
     # @return [Image?] an image that represents a portion of the containing image
     def subimage(left, top, width, height); end
 
+    ##
+    # Returns an image from a binary string of packed RGBA values. (e.g. from (Image#to_blob))
+    #
+    # @param blob [String] a binary string with 1 byte per channel and pixel (RGBA) (width * height * 4)
```
jlnr: I think I might have added that already - but the old image constructor also supported 4 byte per channel for floating point values, `from_blob` should als officially support it as it doesn't cost us much

rdoc/gosu.rb
```diff
@@ -1038,7 +1050,7 @@ def clip_to(x, y, w, h); end
     #
     # @note Because the returned object is not a true image---it's implemented using vertex buffers and is not backed by a texture---there are restrictions on how it can be used.
     #
-    # @note The width and height of the returned object will be the same values you passed to {record}, regardless of the area you draw on. It is important to pass accurate values if you plan on using {Gosu::Image#draw_as_quad} or {Gosu::Image#draw_rot} with the result later.
+    # @note The width and height of the returned object will be the same values you passed to {record}, regardless of the area you draw on. It is important to pass accurate values if you plan on using {Gosu::Image#draw_as_quad} or {Gosu::Image#draw with an angle set) with the result later.
```
jlnr: The `{braces}` are broken now

src/RenderState.hpp
```diff
@@ -10,10 +10,10 @@ struct Gosu::RenderState
     std::shared_ptr<Texture> texture;
     const Transform* transform;
     ClipRect clip_rect;
-    AlphaMode mode;
+    BlendMode mode;
     
     RenderState()
-    : transform(0), mode(AM_DEFAULT)
+    : transform(0), mode(BM_DEFAULT)
```
jlnr: Can you please update this to `transform(nullptr)` while you're touching this line? Or even better, you can initialize the members inline using `=`

jlnr:
A big PR but definitely the right direction 👍 Time for me to finally fix my own PR :)

lib/gosu/compat.rb
Kjarrigan: You're right. Was a bit confused as I need to call draw_rot if any of the 3 is set because draw doesn't support center so I had to set angle to a default as well but I guess now I got it ;-) but not sure how you came to 

> out center_x uses center_y: 0

all defaulted to 0.5

jlnr: It only defaults to 0.5 when `angle` is given. So:

```ruby
image.draw x, y, z, center_x: 0.5 # => center_y = 0.5 (surprising)
image.draw x, y, z => center_y = 0.0 (OK)
```

lib/gosu/compat.rb
Kjarrigan: Ok my girls are sleeping againg.....So - you're right. Performance might really be an issue here as drawing is done for possibly thousands of things. I use Ruby for over 10 years now and it was always fast enough with pure Ruby code or existing gems but I tend to forget that *this* is not my game but the layer below which should be fast. I'll see if I can imitate what you already did with the other methods supporting options.

jlnr:
Regarding my previous comments on `gosu.i`:

> If Gosu::Image works (because it's default constructible now?), we should always just return images directly instead of working with pointers and `%newobject`.

I take that back :) First, in some cases we need to work with pointers because methods can return `nil` in Ruby. Second, when I remove the pointer, it seems that SWIG generally wraps my method in `new Image(...)` to turn `Gosu::Image` into `Gosu::Image*` anyway, so we might as well keep doing it ourselves.

tl;dr `%newobject Gosu::Image*` is fine :) No need to mess with it.

Kjarrigan:
@jlnr 

Now that the screenshot is gone  (#451) most of this PR could be removed I think, because the "tests" relied on it. So basically only keep the Renaming of AlpaMode -> BlendMode and drop everything else (maybe some of the RDoc fixes too)?

Kjarrigan:
Ok. I can live with a not working windows, because some tests are better than none. I'll rework/rebase it later today and report back!

Issue #442, opened 2018-03-16
Drop Ruby < 2.0 (or even < 2.3) Support


Kjarrigan:
Hi @jlnr,

thought about this topic again. You mentioned a few gems that are used with gosu which wouldn't work any longer. Can you tell me what gems you meant - maybe I can try to make them Ruby 2.x compatible. (I haven't really used anything besides gosu so far. I played a bit with chipmunk and have heard of Ashton and Texplay but never used any of those.)

Greeting,
Kjar

jlnr:
Libraries that should be tested on Ruby 2.0 (I think that's a good baseline, some versions of macOS still use it):
- [ ] Texplay
- [ ] Ashton
- [ ] Releasy
- [ ] Maybe even fidgit? It's part of the Gosu org now.

I'm still hopeful that most broken stuff (shaders? Ashton's screenshot functionality?) can be salvaged into the Gosu gem proper. Having one binary gem in the community is enough of a burden IMHO... :(

Libraries that are out of scope:
- OpenGL (not in the Gosu org, has its own maintainers)
- Chipmunk (same)
- Chingu (some of it is already supported by Gosu itself)

jlnr:
I was also thinking about this last night. I'm bundling utf8proc now for sane Unicode handling, and I think it'll add a few hundred KB to every `.so` file on Windows (the gem contains one binary for each Ruby version).

shawn42:
Do feel free to check out chipmunk on the latest versions and let me know if it needs any love. I'm pretty much the only one maintaining it these days.

Kjarrigan:
Thanks to you both. I'll See what I can do. I had a few minutes so I tried Texplay (as asthon has it as dependency) and it seemed to work with 2.4 (the first few examples worked). Just by the stars releasy is the most popular of your list. But I neither have a Windows nor Mac so we'll See how i could test it. 
But die I understand it corectly that merging e.g. asthon directly into gosu is an option?


To the increasing size problem I have no idea how to handle it atm. How do other gems handle this? I'll ask my coworker Lars. He has a lot mode experience in maintaining and deploying gems.

PS: writing on tablet with german autocomplete sucks...

jlnr:
> merging e.g. asthon directly into gosu is an option?

Not a direct "copy and paste" merge :) But Ashton had screenshot functionality and we are already adding that. If Gosu has native texture shader support in the future, that could replace the Ashton gem (but probably with a different interface).

[...]

PR #444, opened 2018-04-04
Drop support for Ruby 1.8 (#442)

jlnr:
The 0.14 release is about better Unicode support, and Unicode was a mess in Ruby 1.8, so let's get rid of it.

@Kjarrigan I was about to drop Ruby 1.9.x support as well, but Travis seems to be stuck on Ubuntu 14.04 (https://github.com/travis-ci/travis-ci/issues/5821). Maybe I'll simply drop support for it on Windows, the 64-bit Windows gem already has special version requirements (≥ Ruby 2.1.0).


jlnr:
Or maybe we can even drop support for <2.3 altogether on Windows? Windows folks don't suffer from outdated system packages anyway.

PR #451, opened 2018-05-14
Synchronize touch & non-touch tutorials

jlnr:
- [ ] Try to make screenshot() more reliable
- [ ] Make sure screenshot() works on HiDPI
- [ ] See if screenshot() works on iOS (OpenGL ES)?

### DIFF
--- a/examples/Tutorial-Touch/Tutorial-Touch.cpp
+++ b/examples/Tutorial-Touch/Tutorial-Touch.cpp
@@ -3,7 +3,6 @@
 #include <cmath>
 #include <cstdlib>
 #include <list>
-#include <memory>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


jlnr:
Previous PR: #431

jlnr:
Also related to HiDPI/LoDPI: #452

jlnr:
Some problems along the way:
- [x] Skip `render{}` tests on Windows as Appveyor doesn't have a complete/modern OpenGL implementation available :(
- [ ] Fix tests on Travis (I probably need to stop using Ruby 1.9.3?)

jlnr:
Hmm. We regularly hear from users that Gosu games run very slowly, and then it turns out that their drivers are somehow borked and don't use hardware acceleration for OpenGL. Maybe Gosu should display a message box on Windows when the renderbuffer extension cannot be found, because that will usually indicate missing graphics drivers?

jlnr:
@Kjarrigan I have removed the `screenshot` method instead of replacing its implementation with `render(window.width, window.height) { window.draw }` (as I had originally planned):
* It's not intuitive that `screenshot` would call `draw` a second time
* More importantly, it's not clear how `screenshot` should handle screens with 200% zoom. Isn't it confusing when `window.screenshot.width == window.width * 2`? 🤔 Now the burden for zooming (with `Gosu.scale`) or not is on the user.
Maybe we can return `screenshot` when I feel more confident in Gosu's HiDPI strategy. The good news is that we'll probably need `glReadPixels` in another place soon – it seems it's the best way to implement `Image#to_blob` on iOS/OpenGL ES, where there's no glGetTexImage.

Kjarrigan:
Hi @jlnr ,

I'm a bit sad about this :cry: but I guess you're right. It just wasn't stable enough, but now we're back to the "old" state were we have no way of testing the graphical features. Why is it so hard :thinking: to get a good and reliable screenshot. Most games offer this feature and they always represent what I see, although they are never compared to the exact same frame on another machine, so they might have the problems as well.

jlnr:
See my comment in #443, we can write tests :) It's just that the screenshot convenience method is gone for now. I'm not even against adding it back if people request it, but preferably only after the remaining HiDPI issues, and after we know if `render` works "in the wild".

PR #457, opened 2018-07-15
Better Font/text interface

jlnr:
This branch builds on #429 and #448:

* [ ] ~~Replace `Font#[]=` by a block passed to the constructor, for example: `Font.new(...) { |char| Image.new("glyphs/#{char}.png") }`.~~; Too exotic, not a priority for 1.0.
* [x] Decide what to do with font's default flags. Right now it is always bold - I think it shouldn't be, but Font.new should accept `bold: true, italic: true` as an argument (breaking change!)
* [x] Merge both versions of `create_text` into a single method
* [x] Add `Font#draw_markup` and `Image.from_markup` as suggested in #304. (In Gosu 1.0, *only* these methods will parse `<b>markup</b>`.)
* [x] Don't forget to adjust motion-gosu as well
* [x] ~~Can we please kill `std::wstring` while we're at it?~~ (no, but limited to WinUtility.hpp)
* [ ] ~~Re-enable unit tests~~ (this will be its own focused PR)
* [ ] ~~Add unit tests for non-markup methods~~


jlnr:
Oops. Accidentally overwrote this with changes from #448

PR #485, opened 2019-01-19
Added Gosu.fill_circle which renders a filled circle using lines

cyberarm:
A naive, work-in-progress, implementation of a filled circle.
This pull request, at this point, is a request for comment on desirable method name(s), parameters, and if a more efficient technique is needed or available (Gosu's internals can be a little confusing to navigate at times.)

I didn't name it `Gosu.draw_circle` as I was considering adding a separate outline only circle named that.
However I could refactor it to include the border/outline and only fill in the circle if it has a non-transparent color. for example: `Gosu.draw_circle(x, y, radius, color, fill_color = transparent, z = 0, alphamode = default)`

screenshot from 2019-01-18 21-58-58

### DIFF
--- a/Gosu/Graphics.hpp
+++ b/Gosu/Graphics.hpp
@@ -80,6 +80,9 @@ namespace Gosu
                               double x2, double y2, Color c2,
                               ZPos z, AlphaMode mode = AM_DEFAULT);
 
+        static void fill_circle(double x, double y, double radius, Color color,
+                                  ZPos z, AlphaMode mode = AM_DEFAULT);
+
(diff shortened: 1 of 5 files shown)


jlnr:
Thanks for the PR, and sorry to be such a downer, but I've decided not to include more geometrical primitives in Gosu. There's too many different ways to draw circles - inner radius, outer radius, color gradients, etc. (I've also blocked line widths in #433.)

I think that each game should implement its own version based on its specific needs, and if there is demand to extract this into a gem, it should be a separate gem built on top of Gosu. Drawing circles in Ruby could be made performant enough using `Gosu.record`.

Issue #478, opened 2018-11-22
Use Ubuntu 16.04 Xenial in CI


jlnr:
Travis supports a more recent version of Ubuntu now. Yay! This also means we can use much more of C++14/17 in Gosu.

https://github.com/travis-ci/travis-ci/issues/5821#issuecomment-440899016

PR #488, opened 2019-03-11
Support Xenial

pusewicz:
Closes https://github.com/gosu/gosu/issues/478

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ addons:
       - cmake
       # For Linux Soundtests (like xvfb for sound: lets us headlessly run audio tests)
       - alsa-base
+      - alsa-utils
 
 services:
   - xvfb
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: cpp
 
 matrix:
   include:
+    # Ubuntu 16.04 Xenial Xerus
+    - os: linux
+      dist: xenial
+      sudo: required
+      compiler: gcc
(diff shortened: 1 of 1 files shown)


pusewicz:
@jlnr I've opened this as a Draft first to see if it builds fine in Travis, but I can see no job being kicked-off? Do I need to do anything special to trigger it?

jlnr:
Regarding the actual PR contents, I think the goal should be to replace the old Trusty base by Xenial, not to supplant it. That way we can start to use newer C++ language constructs.

Not sure why it's not building! As a general heads-up, I would advise you (and also @cyberarm who did another CI-related PR) to keep an eye on how whether Travis still operates as expected, context: https://news.ycombinator.com/item?id=19218036

pusewicz:
@jlnr Yeah, absolutely, I first wanted to add Xenial, to see if it builds at all. I'll see if changing the config file again triggers it.

cyberarm:
@pusewicz Might be able to use pulseaudio's dummy sound output.

https://github.com/travis-ci/travis-ci/issues/9149#issue-291913854
http://manpages.ubuntu.com/manpages/xenial/en/man1/pulseaudio.1.html

pusewicz:
@cyberarm That seems to have made the trick!

jlnr:
Thanks a lot everyone! 🎉🎉🎉 Merging now. @Kjarrigan Feel free to resurrect this ticket if you think going for pulseaudio is the wrong direction here :)

PR #490, opened 2019-03-18
Tweak deprecation warnings so that URLs are accessible

ragesoss:
Putting a space at the end of the URL will let terminals parse the link correctly, instead of opening a bad URL like "https://www.rubydoc.info/github/gosu/gosu/master/Gosu/Image.,"


jlnr:
Oh, makes sense! Thanks. I think that a space before a period links a bit odd though, can you please remove the . altogether, letting the message end with the URL?

ragesoss:
Done.

jlnr:
Thanks!

PR #501, opened 2019-08-22
Implement C API

cyberarm:
Currently images can be loaded and saved out (and destroyed.)
My goal is to emulate the Ruby gem as much as possible. That is, Implement the 9 classes and methods as closely as is practical.

C Image basic load and save test:
```C
#include "../gosu/Gosu/Image.h"

struct Gosu_Image *image;
int main()
{
  image = Gosu_Image_create("/home/cyberarm/Downloads/waterSim.png", 0x00000000);
  Gosu_Image_save(image, "path.png");
  Gosu_Image_destroy(image);
  return 0;
}
```

Edit:
### Current State
Fully Implemented: _(probably)*_
* [X] **Song**
* [X] **Sample**
* [X] **Channel**
* [X] **TextInput**
* [X] **Window**
* [X] **Color**
* [X] **Font**
* [x] **Image**
* [X] **Gosu**
* [X] Button Constants

* [x] Gosu Constants
  * [x] TODO: String constants `VERSION`, `LICENSES` and probably others are returning `NULL` (Might not be initialized at compile time?) https://github.com/gosu/gosu/pull/501#discussion_r323615412
  * [x] TODO: Add various constants for Font alignment and style, AlphaMode (BlendMode), and Image texture modes (smooth, retro, tiled) enums


Gosu/Image.h
```diff
@@ -0,0 +1,46 @@
...
+
+typedef struct Gosu_Image Gosu_Image;
+
+typedef struct Gosu_GLTexInfo {
+  int text_name;
+  double left, right, top, bottom;
+} Gosu_GLTexInfo;
+
+// Constructor
+Gosu_Image* Gosu_Image_create(char *filename, unsigned image_flags);
```
jlnr: Should be `const char*` (for all of the methods)

Gosu/Image.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#ifdef __cplusplus
+  extern "C" {
+#endif
+
+typedef struct Gosu_Image Gosu_Image;
+
+typedef struct Gosu_GLTexInfo {
+  int text_name;
```
jlnr: I think `tex_name` or `texture_name` would be less misleading than `text_name`

Gosu/Gosu.h
```diff
@@ -0,0 +1,57 @@
...
+// #include <Gosu/Channel.h>
+// #include <Gosu/Color.h>
+// #include <Gosu/GLTexInfo.h>
+
+// #include "Gosu/Buttons.h"
+
+// Graphics operations
+void Gosu_gl(*fun_ptr, &func);
+void Gosu_flush();
+Gosu_Image Gosu_render(int width, int height, *fun_ptr &func);
```
jlnr: Should return a pointer

Gosu/Gosu.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Button querying
+bool Gosu_button_down(int id);
+char* Gosu_button_id_to_char(int id);
+int Gosu_button_char_to_id(char* character);
+
+// Misc
+int Gosu_fps();
+char* Gosu_language();
+unsigned Gosu_milliseconds();
```
jlnr: This should be a `long`

Gosu/Gosu.h
```diff
@@ -0,0 +1,57 @@
...
+int Gosu_screen_height();
+
+// Button querying
+bool Gosu_button_down(int id);
+char* Gosu_button_id_to_char(int id);
+int Gosu_button_char_to_id(char* character);
+
+// Misc
+int Gosu_fps();
+char* Gosu_language();
```
jlnr: The functions returning `char*` will be interesting. Probably easiest to keep a `static thread_local std::string` in the wrapper functions so that the returned `const char*` can be converted into a Ruby string without any potential race conditions when multiple threads try to read it. (Not that Gosu itself is thread-safe right now...)

Gosu/Gosu.h
```diff
@@ -0,0 +1,57 @@
...
+void Gosu_draw_line();
+void Gosu_draw_rect(double x, double y, double width, double height, Gosu_Color c, int mode);
+void Gosu_draw_quad();
+void Gosu_draw_triangle();
+
+// Math functions
+double Gosu_angle(double x1, double y1, double x2, double y2);
+double Gosu_distance(double x1, double y1, double x2, double y2);
+double Gosu_angle_diff(double angle1, double angle2);
+double Gosu_offset_x(double theta, double r);
```
jlnr: Please keep the `angle` naming convention

Gosu/Image.h
```diff
@@ -0,0 +1,46 @@
...
+  double left, right, top, bottom;
+} Gosu_GLTexInfo;
+
+// Constructor
+Gosu_Image* Gosu_Image_create(char *filename, unsigned image_flags);
+Gosu_Image* Gosu_Image_create_from_blob(char *source, int width, int height, unsigned image_flags); // support for RMagick blob objects
+Gosu_Image* Gosu_Image_create_from_markup(char *markup, unsigned image_flags);
+Gosu_Image* Gosu_Image_create_from_text(char *text, unsigned image_flags);
+
+// TODO: make this return an array
```
jlnr: Oh...this is going to be interesting :)

Gosu/Image.h
cyberarm: Went with `texture_name` as it seems the most logical

src/ImageWrapper.cpp
```diff
@@ -3,16 +3,85 @@
...
+
+  // Rendering
+  void Gosu_Image_draw(Gosu_Image* image, double x, double y, double z, unsigned color, int mode)
+  {
+    reinterpret_cast<Gosu::Image*>( image )->draw(x, y, z, 1.0, 1.0, color, (Gosu::AlphaMode)mode);
+  }
+
+  void Gosu_Image_draw_rot(Gosu_Image* image, double x, double y, double z,
+                           double angle, double center_x, double center_y,
+                           double scale_x, double scale_y, unsigned color, int mode)
```
jlnr: I think using `unsigned` for enums would be more consistent, especially because every enum might become a bitset over night.

src/ImageWrapper.cpp
```diff
@@ -3,16 +3,85 @@
...
+  {
+    reinterpret_cast<Gosu::Image*>( image )->draw_rot(x, y, z,
```
jlnr: Do you think it would be worth the CPU cycles to check if `image != nullptr`? Although you couldn't really do anything useful but `abort` anyway.

src/ImageWrapper.cpp
```diff
@@ -3,16 +3,85 @@
 extern "C" {
   #include <Gosu/Image.h>
 
+  // Constructors
   Gosu_Image* Gosu_Image_create(char* filename, unsigned image_flags)
```
jlnr: Needs more const :)

src/ImageWrapper.cpp
```diff
@@ -3,16 +3,85 @@
...
+  }
+
+  const char* Gosu_Image_to_blob(Gosu_Image* image)
+  {
+    Gosu::Image* gosu_image = reinterpret_cast<Gosu::Image*>( image );
+
+    return reinterpret_cast<const char*>( gosu_image->data().to_bitmap().data() );
+  }
+
   void Gosu_Image_save(Gosu_Image* image, char* filename)
```
jlnr: More const here too

Gosu/Image.h
```diff
@@ -33,12 +33,16 @@ void Gosu_Image_draw(Gosu_Image *image, double x, double y, double z, unsigned c
...
-void Gosu_Image_draw_as_quad();
+void Gosu_Image_draw_as_quad(Gosu_Image* image, double x1, double y1, unsigned color1,
+                               double x2, double y2, unsigned color2,
+                               double x3, double y3, unsigned color3,
+                               double x4, double y4, unsigned color4,
+                               double z, int mode);
 
 // Operations
 void Gosu_Image_insert(Gosu_Image *image, Gosu_Image *source, int x, int y);
 void Gosu_Image_save(Gosu_Image *image, char *filename);
```
jlnr: I guess `Image_save` would be a good first example of a function that will probably benefit from having some kind of error code, even if it is only success and failure.

jlnr: I wonder if specific error codes would even help, or if it would be easier to just wrap everything that can fail in `try...catch`, store the exception message in a `static thread_local std::string` somewhere, and just return `true/false` from failable methods.

Gosu/Gosu.h
cyberarm: Is this how `static thread_local std::string` should be used for the desired affect?
```C
  const char* Gosu_Image_to_blob(Gosu_Image* image)
  {
    Gosu::Image* gosu_image = reinterpret_cast<Gosu::Image*>( image );
    static thread_local std::string blob = (const char*) gosu_image->data().to_bitmap().data();

    return (const char*)&blob;
  }
```

Gosu/Gosu.h
cyberarm: I believe this is corrected

src/WindowWrapper.cpp
```diff
@@ -1,22 +1,24 @@
...
+#include <Gosu/WindowForWrapper.hpp>
 
 extern "C" {
   #include <Gosu/Window.h>
 
   // Constructor
   Gosu_Window* Gosu_Window_create(int width, int height, int fullscreen, double update_interval, int resizable)
   {
-    return reinterpret_cast<Gosu_Window*>( new Gosu::Window(width, height, fullscreen, update_interval, resizable) );
+    return reinterpret_cast<Gosu_Window*>( new Gosu::WindowForWrapper(width, height, fullscreen, update_interval, resizable) );
```
cyberarm: I'm not sure why but supplementing `Gosu::Window` for `Gosu::WindowForWrapper` causes undefined reference errors in the resulting compiled libgosu.so.

```
/usr/bin/ld: ../gosu/cmake/build/libgosu.so: undefined reference to `Gosu::WindowForWrapper::set_draw(void (*)())'
/usr/bin/ld: ../gosu/cmake/build/libgosu.so: undefined reference to `Gosu::WindowForWrapper::WindowForWrapper(int, int, bool, double, bool)'
/usr/bin/ld: ../gosu/cmake/build/libgosu.so: undefined reference to `Gosu::WindowForWrapper::set_update(void (*)())'
```

Testing with: https://gist.github.com/cyberarm/ef957e30b0e5bb9302d6cbcc4dd892e9

Gosu/Gosu.h
jlnr: Generally yes, but it has to be `return blob.c_str()`.

`to_blob` is probably the only example where it annoys me that this will never free the memory. :/ I guess it's a good start though, maybe we can figure out some way to save space later.

jlnr: For strings, generally yes, but you have to `return blob.c_str()`. And you must separate the definition and initialization/assignment of the `static` variable, because otherwise the variable will only be initialized once. And in this particular case, you should use a `vector` or stick with `Gosu::Bitmap` instead of a `string`, because `to_blob` can contain null bytes, which would terminate the blob early.

```cpp
static thread_local Gosu::Bitmap bitmap;
bitmap = gosu_image->data().to_bitmap();
return reinterpret_cast<const char*>(bitmap.data());
```

`to_blob` is probably the only example where it annoys me that this will never free the memory :/ I mean, you could add a hack, like calling `Gosu_Image_to_blob(nullptr)` to free the static storage once it's been copied to Ruby ...

In an oldschool C library like Win32, usually the caller of the function will provide a memory buffer for the function to write into. Since the caller knows the width and height of the image, this would work in this case.

Gosu/Window.h
```diff
@@ -7,16 +7,32 @@
     {
     struct Callbacks
     {
-      void (*update)();
-      void (*draw)();
+      void (*update)() = nullptr;
```
jlnr: Why do you store them in a nested `struct`?

I also think these callbacks should generally take a `Gosu_Window*` as their first arguments, so that the FFI-generated callbacks can pass it on to the right instance. Or does FFI generate a new stub for every object?

Gosu/Window.h
```diff
@@ -31,11 +47,11 @@ Gosu_Window* Gosu_Window_create(int width, int height, int fullscreen, double up
...
 void Gosu_Window_draw(Gosu_Window* window, void function());
-void Gosu_Window_button_down(Gosu_Window* window, void function());
-void Gosu_Window_button_up(Gosu_Window* window, void function());
-void Gosu_Window_drop(Gosu_Window* window, void function());
-void Gosu_Window_needs_redraw(Gosu_Window* window, void function());
-void Gosu_Window_needs_cursor(Gosu_Window* window, void function());
+void Gosu_Window_button_down(Gosu_Window* window, void function(unsigned btn));
+void Gosu_Window_button_up(Gosu_Window* window, void function(unsigned btn));
+void Gosu_Window_drop(Gosu_Window* window, void function(const char* filename));
+void Gosu_Window_needs_redraw(Gosu_Window* window, int function());
```
jlnr: When compiled as C, `int function()` is a variadic function, equivalent to `int function(...)` in C++. `int function(void)` would be more typesafe.

Gosu/Window.h
```diff
@@ -31,11 +47,11 @@ Gosu_Window* Gosu_Window_create(int width, int height, int fullscreen, double up
 // callbacks
 void Gosu_Window_update(Gosu_Window* window, void function());
 void Gosu_Window_draw(Gosu_Window* window, void function());
-void Gosu_Window_button_down(Gosu_Window* window, void function());
-void Gosu_Window_button_up(Gosu_Window* window, void function());
-void Gosu_Window_drop(Gosu_Window* window, void function());
-void Gosu_Window_needs_redraw(Gosu_Window* window, void function());
-void Gosu_Window_needs_cursor(Gosu_Window* window, void function());
+void Gosu_Window_button_down(Gosu_Window* window, void function(unsigned btn));
```
jlnr: Since these are effectively setters, how about renaming them to `Gosu_Window_set_button_down` or `Gosu_Window_on_button_down` etc.?

src/WindowWrapper.cpp
```diff
@@ -32,6 +57,50 @@ void Gosu::WindowForWrapper::draw()
...
+  if (callbacks.drop != nullptr)
+  {
+    callbacks.drop(filename.c_str());
+  }
+}
+
+bool Gosu::WindowForWrapper::needs_redraw() const
+{
+  if (callbacks.needs_redraw != nullptr)
+  {
```
jlnr: { should be on the same line

Gosu/Image.h
```diff
@@ -25,8 +25,8 @@ Gosu_Image* Gosu_Image_create_from_subimage(Gosu_Image *source, int left, int to
 void Gosu_Image_destroy(Gosu_Image *image);
 
 // Image properties
-int Gosu_Image_width(Gosu_Image *image);
-int Gosu_Image_height(Gosu_Image *image);
+const unsigned Gosu_Image_width(Gosu_Image *image);
```
jlnr: The const here does not make a difference since you cannot assign to a return value -> unsigned or unsigned int is enough.

src/ColorWrapper.cpp
```diff
@@ -0,0 +1,110 @@
+#include <Gosu/Color.hpp>
+#include <Gosu/Color.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+Gosu_Color* Gosu_Color_create(unsigned argb)
+{
+  return reinterpret_cast<Gosu_Color*>( new Gosu::Color(argb) );
```
jlnr: Since Color is really just an `unsigned` internally, I think it would be good if we could export it as such, and avoid extra allocations for these tiny objects.

src/ColorWrapper.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+  return reinterpret_cast<Gosu_Color*>( new Gosu::Color(a, r, g, b) );
+}
+Gosu_Color* Gosu_Color_create_from_hsv(double h, double s, double v)
+{
+  return reinterpret_cast<Gosu_Color*>( &Gosu::Color::from_hsv(h, s, v) );
+}
+Gosu_Color* Gosu_Color_create_from_ahsv(Channel alpha, double h, double s, double v)
+{
+  return reinterpret_cast<Gosu_Color*>( &Gosu::Color::from_ahsv(alpha, h, s, v) );
```
jlnr: This will crash, as `from_ahsv` returns a temporary object, which is deleted as soon as the function returns.

src/ImageWrapper.cpp
```diff
@@ -85,9 +85,10 @@ extern "C" {
   const char* Gosu_Image_to_blob(Gosu_Image* image)
   {
     Gosu::Image* gosu_image = reinterpret_cast<Gosu::Image*>( image );
-    static thread_local std::string blob = (const char*) gosu_image->data().to_bitmap().data();
+    static thread_local Gosu::Bitmap bitmap;
+    bitmap = gosu_image->data().to_bitmap();
 
-    return (const char*)&blob;
+    return reinterpret_cast<const char*>(bitmap.data());
```
jlnr: I think I've mentioned this in another review, but this will not work because the bitmap will probably contain null bytes in its data.

Gosu/Constants.h
```diff
@@ -0,0 +1,248 @@
...
+#include <Gosu/Buttons.hpp>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+const char* Gosu_VERSION  = Gosu::VERSION.c_str();
+const char* Gosu_LICENSES = Gosu::LICENSES.c_str();
+
+unsigned Gosu_KB_RANGE_BEGIN     = Gosu::KB_RANGE_BEGIN;
```
jlnr: This should happen in a .cpp file, not in a .h file. Otherwise, if no .cpp file `#include`s this file, these symbols will not be defined, and if two .cpp files include this header, there will be linker errors about duplicate symbols.

cyberarm: Moved this to a .cpp file.
`Gosu::VERSION` and `Gosu::LICENSES` are returning `null`, and I'm not sure why.

src/ImageWrapper.cpp
```diff
@@ -9,6 +11,55 @@ extern "C" {
...
+    return reinterpret_cast<Gosu_Image*>( new Gosu::Image(bitmap, image_flags) );
+  }
+
+  // TODO: replace use of char* with something that doesn't eat NULL
+  Gosu_Image *Gosu_Image_create_from_blob(const char* blob, int columns, int rows, unsigned image_flags)
+  {
+    std::size_t size = columns * rows * 4;
+    Gosu::Bitmap bitmap(columns, rows, Gosu::Color::NONE);
+
+    if (strlen(blob) == size) {
```
jlnr: strlen will not work, since most images will contain null bytes.

jlnr: Oh, I just saw your TODO comment above. Then I'll stop spamming manual TODO comments :D

cyberarm:
### Current State
Fully Implemented: _(probably)*_
* [X] **Song**
* [X] **Sample**
* [X] **Channel**
* [X] **Color**
* [X] **Font**
* [X] **Gosu**
* [X] Button Constants

Implemented with issues:
* [ ] **Window**
  * TODO: Resolve callback issues https://github.com/gosu/gosu/pull/501#discussion_r317400582
* [ ] **Image**
  * TODO: Fix `#to_blob` and `.from_blob` using `const char*`, replace with a `NULL` friendly type
  * TODO: Fix `#gl_tex_info` causing segfault
* [ ] **TextInput**
  * TODO: Fix `#filter` callback returning `NULL` instead of a string (currently ignoring callback and returning `text` due to this)
  * TODO: Fix weird issue of rapid text changing after entering around ~8 characters, _might_ be related to above

* [ ] Gosu Constants
  * TODO: String constants `VERSION`, `LICENSES` and probably others are returning `NULL` (Might not be initialized at compile time?) https://github.com/gosu/gosu/pull/501#discussion_r323615412
  * TODO: Add various constants for Font alignment and style, AlphaMode (BlendMode), and Image texture modes (smooth, retro, tiled) enums

Gosu/Window.h
cyberarm: Added a `void* data` pointer for callbacks to pass when called (Makes Crystal's Box work) https://github.com/gosu/gosu/pull/501/commits/77a6d5aa93cf0302b4ec04707c3cffc729d77022

How should I store the callbacks? In a struct implemented in the proper place or some other method?

src/TextInputWrapper.cpp
```diff
@@ -0,0 +1,94 @@
...
+  #include <Gosu/TextInput.h>
+
+  Gosu_TextInput* Gosu_TextInput_create()
+  {
+    return reinterpret_cast<Gosu_TextInput*>( new Gosu::TextInputForWrapper() );
+  }
+
+  const char* Gosu_TextInput_text(Gosu_TextInput* text_input)
+  {
+    thread_local std::string string;
```
jlnr: Needs static

src/TextInputWrapper.cpp
```diff
@@ -0,0 +1,94 @@
+#include <Gosu/Gosu.hpp>
+
+namespace Gosu
+{
+  class TextInputForWrapper : public Gosu::TextInput
+  {
+    public:
+      TextInputForWrapper();
+      std::string filter(std::string text) const override;
+      const char* (*filter_callback)(const char*) = nullptr;
```
jlnr: I'm not sure how Ruby or Crystal would manage the lifetime of the `const char*` that their functions must return :/ Maybe you could hack this with more indirection, e.g. TextInputForWrapper::filter calls a callback, and that callback calls  `Gosu_TextInput_complete_filter(@text_input, "filter_result")` with the result, and then control returns into `TextInputForWrapper::filter` where this stored value is used as the C++ return value. (Oof. Does that make sense?!)

src/WindowWrapper.cpp
```diff
@@ -0,0 +1,354 @@
...
+  class WindowForWrapper : public Gosu::Window
+  {
+  struct Callbacks
+  {
+    void (*update)(void *data) = nullptr;
+    void (*draw)(void *data) = nullptr;
+    void (*button_down)(void *data, unsigned btn) = nullptr;
+    void (*button_up)(void *data, unsigned btn) = nullptr;
+    void (*drop)(void *data, const char *filename) = nullptr;
+    bool (*needs_redraw)(void* data) = nullptr; // explicitly set to prevent segfault
```
jlnr: This comment seems a bit random because it is not about `needs_redraw`, but all of these callbacks :)

Gosu/Window.h
jlnr: One idiomatic way in modern C++ would be to use std::function and lambdas to bind the callback & data pointer into a single object:

```cpp
#include <functional>
...
class WindowForWrapper : public Gosu::Window
{
public:
    void button_down(Gosu::Button btn) override
    {
        if (update_callback) update_callback(btn.id());
    }

    // Make these public to get rid of many internal setters?
    std::function<void (unsigned btn)> button_down_callback;
    ...
};

void Gosu_Window_set_button_down(Gosu_Window *window, void function(void *data, unsigned btn), void* data)
{
    reinterpret_cast<Gosu::WindowForWrapper*>( window )->button_down_callback = [=](unsigned btn) { function(data, btn); }
}
```

To be honest, your current solution is probably slightly more efficient than wrapping it in a `std::function`. You can also keep using raw function and void pointers, but I think getting rid of the internal setters would shorten the code somewhat.

src/GosuWrapper.cpp
```diff
@@ -193,6 +199,9 @@ extern "C" {
 
   const char* Gosu_default_font_name()
```
jlnr: Hmm, this is actually supposed to be a constant. I'll make a note that this should either be `Gosu::DEFAULT_FONT_NAME`, or that `Gosu::LICENSE` should actually be `Gosu::license()` — either way it should be consistent, and all of them can be constants in Ruby.

src/TextInputWrapper.cpp
cyberarm: Let's see if I got this straight:
`TextInput::filter` calls `filter_callback(text)` and that callback writes to, say, `TextInput::filter_result` and `TextInput::filter` reads from there instead of directly from `filter_callback()`.

Seems simple enough.

jlnr: Right. Except that the writing to `TextInput::filter_result` will have to happen through another function/method because Crystal cannot save data in a std::string instance variable directly.

cyberarm: Implemented https://github.com/gosu/gosu/pull/501/commits/f390785f66f76dd60097785852d88d5f9850ef45

src/ImageWrapper.cpp
```diff
@@ -59,6 +59,23 @@ extern "C" {
     return reinterpret_cast<Gosu_Image *>(image_data.get() ? new Gosu::Image(std::move(image_data)) : nullptr);
   }
 
+  void Gosu_Image_create_from_tiles(const char* source, int tile_width, int tile_height, Gosu_Image** buffer, int buffer_size, unsigned image_flags)
```
cyberarm: @jlnr I implemented `Gosu_Image_create_from_tiles` by taking in and writing to a buffer, instead of returning an array.

I tried to use a `std::vector` however returning `images.data()` resulted in crashes, I assume were caused by going out of scope; Using the returned pointers would crash when accessing `Gosu::Image` functions. Some of the pointer addresses would be all or nearly all zeros.
If I ran `valgrind ruby ./gosu-examples` and loaded the Tutorial example it would work just fine...

I had something like this:
```cpp
void Gosu_Image_create_from_tiles(const char* source, int tile_width, int tile_height, unsigned image_flags)
{
  std::vector<Gosu::Image> gosu_images = Gosu::load_tiles(filename, width, height, image_flags);
  std::vector<Gosu_Image*> images;

  for (Gosu::Image img : gosu_images)
  {
    images.push_back(reinterpret_cast<Gosu_Image*>( new Gosu::Image(img) );
  }

  return images.data();
}
```

jlnr: Would it maybe be easier to have a function parameter that will be called one for every returned image with a context pointer, which it uses to add them all to the same array? Because your pasted code snippet returns a pointer to local data, and I don't think the updated function signature makes it clear how the number of created images will be returned.

Gosu/Window.h
```diff
@@ -28,7 +28,7 @@ int Gosu_Window_width(Gosu_Window* window);
 int Gosu_Window_height(Gosu_Window* window);
 bool Gosu_Window_fullscreen(Gosu_Window* window);
 double Gosu_Window_update_interval(Gosu_Window* window);
-int Gosu_Window_resizable(Gosu_Window* window);
+bool Gosu_Window_resizable(Gosu_Window* window);
```
jlnr: This is probably a matter of taste, but I think is_resizable would look better in C. C++ often uses methods like `empty()` instead of `is_empty()`, but I'm not sure if this pattern applies to C, where names are more random. `isupper` and friends have an `is` prefix, for what it's worth.

src/ImageWrapper.cpp
cyberarm: So, take in a function pointer that takes an `Gosu_Image*` and call it for each image?
e.g.
```cpp
void Gosu_Image_create_from_tiles(void* callback(Gosu_Image* ptr), const char* source, int tile_width, int tile_height, unsigned image_flags)
{
  std::vector<Gosu::Image> gosu_images = Gosu::load_tiles(filename, width, height, image_flags);

  for (Gosu::Image img : gosu_images)
  {
    callback( (reinterpret_cast<Gosu_Image*>( new Gosu::Image(img) );
  }
}
```

jlnr: Yep, except you probably need a context pointer again for Crystal. And making it a reference in the for loop can be a little more efficient: `for (Gosu::Image& image : gosu_images) ...`

src/ImageWrapper.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // FIXME: accessing gosu_texture_info causes segfault
+    const Gosu::GLTexInfo* gosu_texture_info = reinterpret_cast<Gosu::Image*>( image )->data().gl_tex_info();
+    if (gosu_texture_info) {
+      tex_info->texture_name = gosu_texture_info->tex_name;
+      tex_info->left         = gosu_texture_info->left;
+      tex_info->right        = gosu_texture_info->right;
+      tex_info->top          = gosu_texture_info->top;
+      tex_info->bottom       = gosu_texture_info->bottom;
+      return tex_info;
+    }
```
cyberarm: Can this simply return a pointer to a `Gosu::GLTexInfo` struct cast as a `Gosu_GLTexInfo`, since they share the same layout?

```C
Gosu::Image gosu_image = reinterpret_cast<Gosu::Image*>( image );
Gosu_GLTexInfo* texture_info = reinterpret_cast<Gosu_GLTexInfo*>( new Gosu::GLTexInfo(gosu_image->data().gl_tex_info()) );

return texture_info;
```

jlnr: Should work, even if one is defined inside `extern "C"` and the other is C++. Let's go with the simple solution first.

cyberarm:
Force pushed the `c_library` branch to bring in the changes from #517.

cyberarm:
@jlnr Interesting note, while figuring out how to compile Gosu on Windows through MINGW, I found that if I only compiled Constants.cpp and Version.cpp then `Gosu_VERSION` and `Gosu_LICENSES` had values, but after I added the rest of the files, they were blank/null again.

[Gist with Makefile](https://gist.github.com/cyberarm/805ae7fa3e6046aa20ad8ba7484c2a2f#file-makefile)

jlnr:
@cyberarm [The problem here is that the order of initialization for global variables is undefined](https://stackoverflow.com/questions/3746238/c-global-initialization-order-ignores-dependencies/3746249#3746249), so there's no guarantee that Gosu::VERSION will be initialized before Gosu_VERSION. I think the safest way to solve the issue is to export it as a function.

cyberarm:
@jlnr I believe this is "done" feature-wise, although I haven't added any additional error handling.

I am getting this warning for the Sample wrapper and am not sure how to resolve it:
```
[ 55%] Building CXX object CMakeFiles/gosu.dir/home/cyberarm/Code/gosu/src/SampleWrapper.cpp.o
/home/cyberarm/Code/gosu/src/SampleWrapper.cpp: In function ‘void Gosu_Sample_destroy(Gosu_Sample*)’:
/home/cyberarm/Code/gosu/src/SampleWrapper.cpp:13:53: warning: possible problem detected in invocation of delete operator: [-Wdelete-incomplete]
   13 |   delete( reinterpret_cast<Gosu_Sample *>( sample ) );
      |                                                     ^
/home/cyberarm/Code/gosu/src/SampleWrapper.cpp:13:11: warning: invalid use of incomplete type ‘struct Gosu_Sample’
   13 |   delete( reinterpret_cast<Gosu_Sample *>( sample ) );
      |         ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/cyberarm/Code/gosu/src/SampleWrapper.cpp:2:
/home/cyberarm/Code/gosu/cmake/../Gosu/Sample.h:9:16: note: forward declaration of ‘struct Gosu_Sample’
    9 | typedef struct Gosu_Sample Gosu_Sample;
      |                ^~~~~~~~~~~
/home/cyberarm/Code/gosu/src/SampleWrapper.cpp:13:53: note: neither the destructor nor the class-specific operator delete will be called, even if they are declared when the class is defined
   13 |   delete( reinterpret_cast<Gosu_Sample *>( sample ) );
```

jlnr:
@cyberarm Looks like a simple typo — you should reinterpret_cast to Gosu::Sample, not Gosu_Sample :)

cyberarm:
@jlnr Ah, well that's an easy fix! 🙃
Anything else that should be done before this is merged?

src/GosuWrapper.cpp
```diff
@@ -0,0 +1,233 @@
...
+    std::function<void (void)> callback;
+    callback = [=](void) { function(data); };
+
+    Gosu::Graphics::gl(callback);
+  }
+
+  Gosu_Image *Gosu_render(int width, int height, void function(void *data), void *data, unsigned image_flags)
+  {
+    std::function<void (void)> callback;
+    callback = [=](void) { function(data); };
```
jlnr: Using `(void)` to denote an empty parameter list instead of `()` is a C-ism and deprecated in C++. Even though this is an `extern "C"` area, this code is compiled as C++, and doesn't need the voids.

cyberarm: :book: Ah, interesting.
:heavy_check_mark: should be corrected now.

cyberarm:
Hey, @jlnr. Have you had time to review this further?

jlnr:
@cyberarm I'm afraid not yet :( Crunch time for my current project has just been extended throughout this month.

cyberarm:
@jlnr Oof, be careful not to burn out. :slightly_smiling_face:

jlnr:
@cyberarm No need to worry I think! This job actually makes me want to get back into hobby coding even more :D

src/WindowWrapper.cpp
```diff
@@ -0,0 +1,341 @@
...
+    void set_button_down(void function(void *data, unsigned btn), void *data);
+    void set_button_up(void function(void *data, unsigned btn), void *data);
+    void set_drop(void function(void *data, const char *filename), void *data);
+    void set_needs_redraw(bool function(void *data), void *data);
+    void set_needs_cursor(bool function(void *data), void *data);
+    void set_close(void function(void *data), void *data);
+
+    void update() override;
+    void draw() override;
+    void gosu_button_down(unsigned btn); // Enables fullscreen toggle
```
jlnr, line 20: Hmmm, maybe clearer to name this handle_fullscreen_shortcuts or default_button_down?

src/WindowWrapper.cpp
```diff
@@ -0,0 +1,341 @@
+#include <Gosu/Gosu.hpp>
+
+namespace Gosu
+{
+  class WindowForWrapper : public Gosu::Window
```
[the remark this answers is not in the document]
jlnr: If not, I can also give it a try. I've been meaning to import Gosu into CLion anyway, after using it every day at work...

src/TextInputWrapper.cpp
```diff
@@ -0,0 +1,100 @@
...
+      std::string filter(std::string text) const override;
+      std::function<void (const char* text)> filter_callback;
+      std::string filter_result = "";
+  };
+}
+
+Gosu::TextInputForWrapper::TextInputForWrapper() : Gosu::TextInput()
+{
+}
+std::string Gosu::TextInputForWrapper::filter(std::string text) const
```
jlnr, line 18: Missing newline

src/TextInputWrapper.cpp
```diff
@@ -0,0 +1,100 @@
...
+  #include <Gosu/TextInput.h>
+
+  Gosu_TextInput* Gosu_TextInput_create()
+  {
+    return reinterpret_cast<Gosu_TextInput*>( new Gosu::TextInputForWrapper() );
+  }
+
+  const char* Gosu_TextInput_text(Gosu_TextInput* text_input)
+  {
+    static thread_local std::string string;
```
jlnr, line 40: `static thread_local` is redundant: https://stackoverflow.com/a/22794640

Can you please remove the `static`?

Gosu/Gosu.h
```diff
@@ -0,0 +1,79 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <Gosu/Window.h>
+#include <Gosu/Image.h>
+#include <Gosu/Font.h>
+#include <Gosu/Color.h>
+#include <Gosu/TextInput.h>
+// #include <Gosu/Sample.h>
```
jlnr, line 12: Why are these commented out?

cyberarm: If I remember correctly, I was toying with using the C API in C and those headers weren't implemented yet.
I'll uncomment them in a bit.

src/WindowWrapper.cpp
```diff
@@ -0,0 +1,341 @@
...
+    std::function<void ()> close_callback;
+  };
+}
+
+Gosu::WindowForWrapper::WindowForWrapper(int width, int height, bool fullscreen,
+                                         double update_interval, bool resizable)
+: Gosu::Window(width, height, fullscreen, update_interval, resizable) {
+}
+
+void Gosu::WindowForWrapper::set_update(void function(void *data), void *data)
```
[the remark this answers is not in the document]
cyberarm: Can FFI directly set these variables?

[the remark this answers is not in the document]
jlnr: No, but Gosu_Window_set_update can, getting rid of one level of indirection.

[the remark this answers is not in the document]
cyberarm: Ah! Makes sense. :upside_down_face: 
Changed.

PR #504, opened 2019-08-28
Audio Playback Position Duration Sample Rate

edwinacunav:
This time I'll try to submit my changes to the audio department the right way. The main changes would be the addition of playback position and duration and enabling functions like resume because I do think it's useful just to call it instead of passing a boolean to play(boolean) once again. (Especially useful if you want a single button to let you pause and resume songs at will.) With resume we no longer need to keep track of the initial settings. About position I might say it's convenient if anybody wants to show how many seconds are left before the song ends or simply tell the player how long he's been listening to the same song. I know for sure it gets reset after a new song starts playing. (I hate looping songs because they remind me of an closed source game engine...) Keep in mind it'll be hard or even impossible to keep track of a position on songs with variable bitrates. :S

From Ruby those things should be called:

```ruby
@song.position # Returns a Float
@song.position = Float # in seconds, not frames nor samples, just seconds
@song.position_minutes
@song.position_hours # Just in case it's some instrumental soundtrack like a symphony of sorts
@song.duration
@song.resume
```
I've been using it while playing OGG files so I know they run smoothly.


Gosu/Buttons.hpp
```diff
@@ -51,6 +51,9 @@ namespace Gosu
         KB_DOWN            = 81,
         KB_HOME            = 74,
         KB_END             = 77,
+        KB_PRINTSCREEN     = 70,
+        KB_SCROLLLOCK      = 71,
+        KB_PAUSE           = 72,
```
jlnr, line 56: Thanks.

Gosu/Bitmap.hpp
```diff
@@ -88,6 +91,8 @@ namespace Gosu
         }
     };
     
+    void enable_flip_y(bool flip_y);
+    void load_image_inverse_color(Bitmap& bitmap, const std::string& filename);
```
jlnr, line 95: These two features seem too obscure to add them to Gosu, they'd be great use cases for a general-purpose shader solution though.

Gosu/Audio.hpp
```diff
@@ -136,6 +136,20 @@ namespace Gosu
...
+        //! Returns the current song's playback position in hours
+        int position_hours();
+        //! \param pos position can be any value from 0.0 through the last second
+        void set_position(float pos);
+        //! Returns how long your song might be
+        float duration() const;
+        //! Returns a string with the audio format: Mono or Stereo
+        const char* format();
+        //! The song's sample rate aka samples per second
+        int sample_rate() const;
```
jlnr, line 152: The format and sample rate are implementation details, and they're not accurate because I'm overriding them in both MPEGFile and AudioToolboxFile.

Gosu/Audio.hpp
```diff
@@ -136,6 +136,20 @@ namespace Gosu
...
+        //! Returns the current song's playback position in seconds
+        int position() const;
+        //! Returns the current song's playback position in minutes
+        int position_minutes();
+        //! Returns the current song's playback position in hours
+        int position_hours();
+        //! \param pos position can be any value from 0.0 through the last second
+        void set_position(float pos);
+        //! Returns how long your song might be
+        float duration() const;
```
jlnr, line 148: Having too many representations for times (int seconds, unsigned long Gosu::milliseconds, float duration) is dangerous because we don't have any type safety here or in Ruby.
In my opinion, having a floating-point number of seconds is the best approach. However it should be double, not float, for consistency with `double volume()` and all of the rendering APIs, which use doubles.

src/Audio.cpp
```diff
@@ -141,7 +141,7 @@ class Gosu::Song::BaseData
     double volume_;
 
 protected:
-    BaseData() : volume_(1) {}
+    BaseData() : volume_(1), frames(0) {}
```
jlnr, line 144: Variable not found

src/Audio.cpp
```diff
@@ -354,6 +359,40 @@ class Gosu::Song::StreamData : public BaseData
         alGetSourcei(al_source_for_songs(), AL_SOURCE_STATE, &state);
         return state == AL_PAUSED;
     }
+
+    float position() const override
+    {
+        float pos;
+        alGetSourcef(al_source_for_songs(), AL_SEC_OFFSET, &pos);
+        return static_cast<float>(frames) / file->sample_rate() + pos;
+    }
```
jlnr, line 368: This method uses a mixture of frames and AL_SEC_OFFSET, and doesn't compile.

src/Audio.cpp
```diff
@@ -366,6 +405,12 @@ class Gosu::Song::StreamData : public BaseData
         alGetSourcei(source, AL_BUFFERS_PROCESSED, &processed);
         for (int i = 0; i < processed; ++i) {
             alSourceUnqueueBuffers(source, 1, &buffer);
+            // Retrieve bits, size and channels
+            alGetBufferi(buffer, AL_BITS, &bits);
+            alGetBufferi(buffer, AL_SIZE, &size);
+            alGetBufferi(buffer, AL_CHANNELS, &channels);
```
jlnr, line 411: Unknown identifiers: bits, size, channels

src/OggFile.hpp
```diff
@@ -88,5 +88,15 @@ namespace Gosu
...
         }
+
+        float duration() override
+        {
+            stb_vorbis_stream_length_in_seconds(stream_);
+        }
+
+        void seek_pos(float pos)
+        {
+            stb_vorbis_seek_frame(stream_, pos);
```
jlnr, line 99: This looks like it's missing a multiplication with the sample rate.

Issue #506, opened 2019-09-11
Add support for setting window icon


cyberarm:
ref: https://wiki.libsdl.org/SDL_SetWindowIcon

jlnr:
My understanding is that we don't need to use SDL_SetWindowIcon to our games an icon on any major operating system, and that it is only necessary for _dynamic_ icon changes while the game is running. Closing this, but if I'm wrong we should resurrect this issue and #507 along with it.

PR #507, opened 2019-09-11
Add `void Window::set_window_icon(const Bitmap&)`

pmer:
Here's a simple implementation for #506. This PR also sets the icon for TutorialExample to Starfighter.bmp.

I've tested it on macOS. I'm not sure what would happen on other platforms, but we should be able to look at SDL_SetWindowIcon() to see what would happen.

Looks like it supports the alpha channel. Yay! :)

Here's a screenshot of TutorialExample. You can see the icon in the dock.

Window icon in TutorialExample on macOS

### DIFF
--- a/examples/Tutorial/Tutorial.cpp
+++ b/examples/Tutorial/Tutorial.cpp
@@ -147,11 +147,9 @@ public:
     {
         set_caption("Gosu Tutorial Game");
 
-        {
-            Gosu::Bitmap icon;
-            Gosu::load_image_file(icon, Gosu::resource_prefix() + "media/Starfighter.bmp");
-            set_window_icon(icon);
-        }
+        Gosu::Bitmap icon;
+        Gosu::load_image_file(icon, Gosu::resource_prefix() + "media/Starfighter.bmp");
+        set_window_icon(icon);
 
         std::string filename = Gosu::resource_prefix() + "media/Space.png";
         background_image = Gosu::Image(filename, Gosu::IF_TILEABLE);
--- a/examples/Tutorial/Tutorial.cpp
+++ b/examples/Tutorial/Tutorial.cpp
@@ -147,6 +147,12 @@ public:
     {
         set_caption("Gosu Tutorial Game");
 
+        {
+            Gosu::Bitmap icon;
+            Gosu::load_image_file(icon, Gosu::resource_prefix() + "media/Starfighter.bmp");
+            set_window_icon(icon);
+        }
+
         std::string filename = Gosu::resource_prefix() + "media/Space.png";
         background_image = Gosu::Image(filename, Gosu::IF_TILEABLE);
 
--- a/src/Window.cpp
+++ b/src/Window.cpp
@@ -247,6 +247,38 @@ void Gosu::Window::set_caption(const string& caption)
     SDL_SetWindowTitle(shared_window(), caption.c_str());
 }
 
+void Gosu::Window::set_window_icon(const Bitmap& icon)
+{
+    #ifdef GOSU_IS_LITTLE_ENDIAN
+    enum { RED_OFFSET = 0, GREEN_OFFSET = 8, BLUE_OFFSET = 16, ALPHA_OFFSET = 24 };
+    #else
+    enum { RED_OFFSET = 24, GREEN_OFFSET = 16, BLUE_OFFSET = 8, ALPHA_OFFSET = 0 };
+    #endif
+
+    void* pixels = const_cast<Color*>(icon.data());
+    int width = icon.width();
+    int height = icon.height();
+    int depth = 8 * sizeof(Color); // Number of bits.
+    int pitch = width * sizeof(Color); // Number of bytes.
+    Uint32 red_mask = 0xff << RED_OFFSET;
+    Uint32 green_mask = 0xff << GREEN_OFFSET;
+    Uint32 blue_mask = 0xff << BLUE_OFFSET;
+    Uint32 alpha_mask = 0xff << ALPHA_OFFSET;
+
+    SDL_Surface* surface = SDL_CreateRGBSurfaceFrom(pixels, width, height, depth, pitch,
+        red_mask, green_mask, blue_mask, alpha_mask);
+    if (surface == nullptr) {
+        throw_sdl_error("Could not create icon");
+    }
+
+    // Surface's pixels are copied.
+    SDL_SetWindowIcon(shared_window(), surface);
+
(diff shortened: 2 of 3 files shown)


pmer:
Window icon in the macOS app switcher

examples/Tutorial/Tutorial.cpp
```diff
@@ -147,6 +147,12 @@ class GameWindow : public Gosu::Window
     {
         set_caption("Gosu Tutorial Game");
 
+        {
+            Gosu::Bitmap icon;
```
jlnr: I'm not sure if this is leftover test code, but I think it seems reasonable enough to have this in the Tutorial. But it should not have the extra scope nesting :)

src/Window.cpp
```diff
@@ -247,6 +247,38 @@ void Gosu::Window::set_caption(const string& caption)
...
+    enum { RED_OFFSET = 0, GREEN_OFFSET = 8, BLUE_OFFSET = 16, ALPHA_OFFSET = 24 };
+    #else
+    enum { RED_OFFSET = 24, GREEN_OFFSET = 16, BLUE_OFFSET = 8, ALPHA_OFFSET = 0 };
+    #endif
+
+    void* pixels = const_cast<Color*>(icon.data());
+    int width = icon.width();
+    int height = icon.height();
+    int depth = 8 * sizeof(Color); // Number of bits.
+    int pitch = width * sizeof(Color); // Number of bytes.
```
jlnr: + ...per row

pmer:
What is the thought about the potential Ruby wrapper for it?

For those of us who have not participated in a Ludum Dare—is it easy to add an icon to a .exe/.app/.desktop for your submission there? Like with a Ruby bundler of some sort, etc? :)

And my first thought about .desktop files is that they have to be installed into the right folder or they don't work, which is different from an .exe or .app. A player couldn't just extract a tgz of a game into their Downloads folder and run the binary—which I've seen a few indie Linux games do... Maybe that is becoming less common, though, and people are moving to MojoSetup and such, which does install a .desktop file correctly.

jlnr:
Yes, [Releasy](https://github.com/gosu/releasy) supports icon files when building a Windows executable (via Ocra). The situation is similar on macOS. The file formats are a bit annoying, but I think this would be easy to work around with Ruby libraries during build time.

I have no idea about Linux, though, and my Linux/Windows workstation decided to die last night 😅 I'll play around some more once the SSD is replaced.

pmer:
Sorry to hear about that. :( Hope you have backups!

pmer:
This PR may be more useful for C and C++ users of Gosu.

edwinacunav:
This can be solved with some other SDL2 functions but we gotta know if there's a way to convert a stb_image array into a SDL_Surface. I know SDL2 introduced SDL_Texture but as you might have noticed by now, SDL_SetWindowIcon accepts a SDL_Surface only...

pmer:
Since edwinacunav pinged this thread and it made me look at it again, I thought of a new question:

For .exe files on Windows, if one sets the icon for the compiled .exe file, will Windows apply that automatically to an SDL2 window or will it be blank until you call SDL_SetWindowIcon()?

pmer:
edwardacunav, I attached a piece of code to this PR that is able to convert a Gosu::Bitmap to an SDL_Surface, and you can get a Bitmap by calling Gosu::load_image_file("filename.png"). I was hoping that might be sufficient.

edwinacunav:
As an experiment, I used your code to save a screenshot in BitmapIO.cpp by adding SDL_image.h as a dependency. It made a better job at saving the PNG image, with some compression, than stb_image.h that failed to do so if you used any compression level above 0... I saved around 33% of disk space! O_O

```
#ifdef GOSU_IS_LITTLE_ENDIAN
enum { RED_OFFSET = 0, GREEN_OFFSET = 8, BLUE_OFFSET = 16, ALPHA_OFFSET = 24 };
#else
enum { RED_OFFSET = 24, GREEN_OFFSET = 16, BLUE_OFFSET = 8, ALPHA_OFFSET = 0 };
#endif

void Gosu::save_screenshot(const Gosu::Bitmap& bitmap, const string& filename)
{
    void *pixels = const_cast<Gosu::Color*>(bitmap.data());
    int depth = 8 * sizeof(Gosu::Color); // Number of bits per pixel.
    int pitch = bitmap.width() * sizeof(Gosu::Color); // Number of bytes per row of pixels.
    uint32_t red = 0xff << RED_OFFSET;
    uint32_t green = 0xff << GREEN_OFFSET;
    uint32_t blue = 0xff << BLUE_OFFSET;
    uint32_t alpha = 0xff << ALPHA_OFFSET;
    SDL_Surface *surf = SDL_CreateRGBSurfaceFrom(pixels, bitmap.width(),
        bitmap.height(), depth, pitch, red, green, blue, alpha);
    if (!surf) {
        SDL_FreeSurface(surf);
        throw runtime_error("Could not retrieve any SDL surface!");
    }
    if (IMG_SavePNG(surf, filename.c_str()) != 0)
        throw runtime_error("Could not save image data to file: " + filename);
    SDL_FreeSurface(surf);
}
```

This is the SWIG like code I added to RubyGosu.cxx to let Ruby call the C++ function.
From Ruby you call:
Gosu.save_screenshot(@rendered_image, "filename.png")

```
static VALUE gosu_save_screenshot(VALUE self, VALUE image, VALUE name)
{
    void *argp1 = 0;
    int res = SWIG_ConvertPtr(image, &argp1, SWIGTYPE_p_Gosu__Image, 0 | 0);
    if (!SWIG_IsOK(res))
        rb_raise(rb_eTypeError, "argument 1 should be a Gosu::Image object");
    std::string filename = RSTRING_PTR(name);
    Gosu::Image *img = reinterpret_cast<Gosu::Image*>(argp1);
    if (!img)
        rb_raise(rb_eRuntimeError, "Failed to reinterpret casting");
    Gosu::save_screenshot(img->data().to_bitmap(), filename);
    return Qnil;
}
//Dozens of lines of code later...
rb_define_module_function(mGosu, "save_screenshot", VALUEFUNC(gosu_save_screenshot), 2);
```

Of course, I had to add -lSDL2_image to the libs list found in the Makefile.

pmer:
Rebasing to force a rebuild on TravisCI. (The original build failed due to SDL 2.0.10 not supporting OS X 10.11.)

pmer:
@edwinacunav That's pretty cool. :)

As for the Windows icon question I had, if I get some time this week I might try to find a answer using a Windows desktop I have lying around the house...

PR #521, opened 2020-01-20
Update bundled libraries

jlnr:


edwinacunav:
If you updated SWIG on Windows and some other Linux distros I wouldn't mind, but on Kubuntu 18.04.3 I know I can only get SWIG version 3.0... Let's see if it gets updated on Kubuntu 20.04...

jlnr:
@edwinacunav Anything below SWIG 4.0.2 (which isn't even out) doesn't work with Ruby 2.7, so I had to install SWIG from source :(

Issue #412, opened 2017-09-03
Analog gamepad input support


jtuttle:
It would be great to be able to read input from the analog sticks on a gamepad (-1.0 ... +1.0) for finer control over directional input.

jlnr:
This topic came up in #gosu, @shawn42 noted that it could be as simple as providing a callback `axis_motion(id, value)` in `Window`. (Not sure about the name yet)

cyberarm:
Should this callback have 3 arguments, i.e. `axis_motion(gamepad_id, axis_id, value)`?

Which SDL Event should be used to capture input: [SDL_ControllerAxisEvent](https://wiki.libsdl.org/SDL_ControllerAxisEvent), [SDL_JoyAxisEvent](https://wiki.libsdl.org/SDL_JoyAxisEvent), or both?

https://wiki.libsdl.org/SDL_Event

jlnr:
Gosu uses the SDL_Controller API when available for a device, SDL_Joy otherwise.

Right now, buttons have different constants for different gamepads: GP_4_BUTTON_4, so it would be more consistent to have GP_4_AXIS_X or something. Redesigning everything to be less C-ish and more object-oriented is an orthogonal task.

cyberarm:
Closing this since [Gosu.axis](https://www.rubydoc.info/gems/gosu/Gosu%2Eaxis) was added to Gosu.

Issue #467, opened 2018-10-04
Distinguish between different sticks and pads on gamepads


taras-dashkevich:
Currently left pad and all sticks using same constants:
- Gosu::GpLeft
- Gosu::GpRight
- Gosu::GpUp
- Gosu::Down

Would be nice to have separately and for sticks additional method  returns vluae (for example -1..1)

Thanks

jlnr:
I've renamed this ticket because a feature request for analog stick inputs already exists (#412). The issue in this ticket is that all sticks and pads are mapped to the same four constants. It's likely that both issues will be closed with the same patch, though.

Tetssuwan:
Thanks @jlnr

jlnr:
Fixed in #524

PR #524, opened 2020-04-30
Refactor Gosu::Input to support gamepad axes and to be able to handle adding/removing controllers at runtime

cyberarm:
- [x] Refactor static gamepad initialization to use `SDL_{CONTROLLER/JOYSTICK}_ADDED /_REMOVED` events

- [x] Store gamepad axes positions from `SDL_{CONTROLLER/JOYSTICK}AXISMOTION` events

- [x] Return axis in range `-1.0..1.0` instead of +/- 32767

- [x] Add `Gosu::Input::axis(gp_axis_id)` (`Gosu.axis`) to get last known axis position, or default of `0.0`

- [x] ~~Add `Gosu::Window::axis_motion(gp_axis_id, value)` (`Gosu::Window#axis_motion`) callback.~~

- [x] Add constants for gamepad axes in the form: `GP_0_AXIS_0` or `GP_0_AXIS_LEFT_X`

- [x] Add a callback for when a controller is added or removed: `Gosu::Window::controller_changed(gp_index_id, connected)` (`Gosu::Window#controller_changed`)
ADDED as `controller_connected` and `controller_disconnected`

- [x] Update `poll_gamepads()` to use the gamepad slot index instead of the game_controller/joystick index for queuing button events

- [x] Rename gamepad axes to `GP_LEFT_STICK_X_AXIS`
- [x] Move `GP_DIRECTION` constants and add `GP_DPAD_DIRECTION` constants

- [x] **Fix edge case where the wrong gamepad slot is freed**

- [x] Added documentation comments for `Gosu::Input::axis` and other added functions
- [x] Remove debug print statements

Addresses https://github.com/gosu/gosu/issues/412

[Filter whitespace changes](/gosu/gosu/pull/524/files?w=1)


src/Input.cpp
```diff
@@ -153,24 +142,61 @@ struct Gosu::Input::Impl
                 }
                 break;
             }
+            case SDL_JOYAXISMOTION: {
+                if (SDL_IsGameController(e->jaxis.which)) {
+                    break; // Ignore as triggers are reported -1..1 instead of 0..1
+                } else {
```
pmer, line 148: Looks like rake doesn't like this line and wants the `else` to be on its own line after the `}`, which is causing the CI to fail. Not sure why rake is so stingy about formatting, but I guess it just is?

cyberarm: I keep forgetting to run the formatter... corrected now. :slightly_smiling_face:

cyberarm:
@jlnr Most everything is implemented, there are a few bugs and edge cases to work out but should be complete enough to review.
I noticed that issue #467 is referenced to #412, should #467 be addressed by this pull request? Should the `GP_DIRECTION` constants be changed to only be the d-pad?

jlnr:
One thing that I'll need to look at is how this behaves when gamepads are added and removed - do existing gamepads keep their indices? If player 1 has to change the batteries in their gamepad, player 2 and 3 should keep their indices (player numbers) throughout the process. Would be even better if the player numbers on the Xbox 360 gamepads would match the index in Gosu.
Hopefully SDL 2 does the hard work for us here 😬

Gosu/Input.hpp
```diff
@@ -92,29 +94,34 @@ namespace Gosu
...
-        
+
         //! Assignable events that are called by update. You can bind these to your own functions.
         //! If you use the Window class, it will assign these to its own methods.
         std::function<void (Button)> on_button_down, on_button_up;
-        
+
+        std::function<void (Button, double)> on_axis_motion;
+
+        std::function<void (int, std::string)> on_gamepad_connected;
```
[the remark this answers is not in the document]
cyberarm: I added `gamepad_name(id)`. It works from within the `gamepad_disconnected` callback since the slot isn't cleared until after the callback is called.

I also added the `button_name(id)` function, I assume its purpose is to return the name for the id? `button_name(Gosu::KbSpace) => "Space"` (TODO: return in lowercase)

jlnr:
@cyberarm I've force-pushed this branch so it contains the latest changes from master. I did actually receive an error message when I did it, but it seems to have worked nevertheless. Reviewing now...

jlnr:
I have added this piece of code to Window::update in Tutorial.cpp:

```cppp
        static std::vector<std::string> gamepadNames;
        std::vector<std::string> newGamepadNames;
        for (int i = 0; i < 4; ++i) {
            newGamepadNames.emplace_back(input().gamepad_name(i));
        }
        if (newGamepadNames != gamepadNames) {
            std::cout << "Gamepads:\n";
            for (int i = 0; i < 4; ++i) {
                std::cout << " [" << i << "]='" << newGamepadNames[i] << "'\n";
            }
            gamepadNames = newGamepadNames;
        }
```

If I start with two gamepads:

```
Gamepads:
 [0]='Jess Tech Dual Analog Pad'
 [1]='X360 Controller'
 [2]=''
 [3]=''
```

...then unplug one of them, I get into an erroneous state already:

```
Gamepads:
 [0]=''
 [1]=''
 [2]=''
 [3]=''
```

jlnr:
Pushing a commit to fix some warnings now. By the way, Xcode/clang++ helpfully shows this:

Bildschirmfoto 2020-06-13 um 13 43 33

This doesn't work because it's a definition, and everything to the right if `=` is part of the expression. However, C++17 allows the special `if (definition; condition)` syntax which I'll use to resolve this.

src/Input.cpp
```diff
@@ -210,63 +339,82 @@ struct Gosu::Input::Impl
...
 private:
     // For button down event: Button name value (>= 0)
     // For button up event: ~Button name value (< 0)
-    vector<int> event_queue;
+    vector<InputEvent> event_queue;
 
     void enqueue_event(int id, bool down)
     {
-        event_queue.push_back(down ? id : ~id);
+        InputEvent s = { down ? id : ~id, InputEventType::ButtonEvent };
```
jlnr, line 382: The silly id/~id trick was just because I was too lazy to create a new struct instead of using `int`s. Now that there's a struct, we should make this more explicit, for example by creating a new InputEventType.

src/Input.cpp
```diff
@@ -153,40 +154,168 @@ struct Gosu::Input::Impl
...
+                        joysticks.push_back(joystick);
+                        gamepad_slot = available_gamepad_slot_index();
+                        gamepad_name = SDL_JoystickNameForIndex(e->jdevice.which);
+                        device_instance_id = SDL_JoystickInstanceID(joystick);
+                    }
+                }
+                if (gamepad_slot >= 0 && device_instance_id >= 0) {
+                    gamepad_slots[gamepad_slot] = device_instance_id;
+                    printf("Gamepad using slot: %i (%s|%i)\n", gamepad_slot, SDL_GameControllerName(SDL_GameControllerFromInstanceID(device_instance_id)), device_instance_id);
+                    enqueue_gamepad_connection_event(gamepad_slot, true, -1);
```
jlnr, line 208: Why no instance_id?

src/Input.cpp
```diff
@@ -153,40 +154,168 @@ struct Gosu::Input::Impl
...
+            }
+            case SDL_CONTROLLERAXISMOTION: {
+                store_axis_value(ButtonName::GP_LEFT_STICK_X_AXIS + e->caxis.axis, (double)e->caxis.value);
+
+                if (int i = gamepad_slot_index(e->caxis.which); i >= 0) {
+                    store_axis_value(ButtonName::GP_0_LEFT_STICK_X_AXIS + e->caxis.axis + (6 * i), (double)e->caxis.value);
+                }
+                break;
+            }
+            case SDL_JOYDEVICEADDED: {
```
jlnr, line 179: Hmm...not sure if this should happen in the event loop. Removed devices are enqueued, but added devices are handled immediately, this means the order of operations can swap. If through some glitch the device is removed and added in the same Input::update() call, it will temporarily exist in the device array twice.

I think the code would be easier to follow if the array modifications here would also happen in Input::update.

jlnr:
AppVeyor is red because Gosu 0.15.x used a Visual Studio version which doesn't support this C++17 syntax yet. I say let's ignore it, Gosu 1.0 will be compiled using a more recent version.

Gosu/Buttons.hpp
```diff
@@ -232,16 +232,77 @@ namespace Gosu
...
+        GP_2_RIGHT,
+        GP_2_UP,
+        GP_2_DOWN,
+
+        GP_3_LEFT,
+        GP_3_RIGHT,
+        GP_3_UP,
+        GP_3_DOWN,
+
+        GP_RANGE_END = GP_3_BUTTON_15,
```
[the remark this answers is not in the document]
jlnr: Or...should it? I need to double-check how GP_RANGE_END is being used.

src/Input.cpp
jlnr: Answer: Because it's not necessary when connecting, only for disconnecting.

jlnr:
I've refactored InputEvent a bit, will look into the remaining issue with the index management tonight or tomorrow. I've also noticed that I cannot turn the ship left in the Tutorial game which checks for GP_LEFT, no matter which stick or D-Pad I use. Maybe I've missed one.

jlnr:
The button ID math got really ugly because each gamepad's buttons are not contiguous anymore. However, I think the last commit fixes all the issues I've seen with GP_LEFT not being calculcated correctly. I've tested this with one SDL_GameController and one SDL_Joystick attached.

Gosu/Input.hpp
```diff
@@ -64,26 +64,28 @@ namespace Gosu
...
         //! encoded string), or NO_BUTTON.
         static Button char_to_id(std::string ch);
-        
+
         //! Returns true if a button is currently pressed.
         //! Updated every tick.
         static bool down(Button btn);
-        
+
+        static double axis(Button btn);
```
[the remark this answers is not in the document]
jlnr: ...which is also what is being returned here according to the SDL docs :)

PR #527, opened 2020-05-05
Song/Sample#length [on hold]

jlnr:
* Cherry-pick Song#duration as Song#length/Sample#length from #504 
* Cherry-pick new button constants from #504
* Remove the AVAudioPlayer implementation of Gosu::Song on iOS
  * This was important on the iPhone 3G because it used hardware acceleration, but modern iPhones are crazy fast anyway.
  * This jungle of audio libraries is terrible. It might be good to drop AudioToolbox in favor of libmpg123 and libsndfile on macOS too. Not sure what to do about iOS though, cross-compiling these libraries sounds annoying.
* Rename data/SampleData and data/StreamData to the more idiomatic pimpl/Impl.
* Add Gosu.seconds, which returns fractional seconds, something that matches Song#length a bit better. It'd be strange if there was only Gosu.milliseconds, but then Song#length returned seconds.


jlnr:
@edwinacunav @Kjarrigan I haven't been able to test this on Windows or Linux yet, feel free to give it a spin and see if the length is reported correctly.

pmer:
I would suggest SDL_Mixer except it doesn't support changing audio playback speed, which is part of Gosu's public API.

cyberarm:
@jlnr @pmer Would SDL2_Sound be helpful in simplify the audio system?
https://icculus.org/SDL_sound
https://hg.icculus.org/icculus/SDL_sound/file/tip/docs/README.txt

jlnr:
Changing audio playback speeds is one of my pet features :( Just adding ±5% speed adds so much variation to repeating sounds. And SDL_mixer has the distinction between samples and songs that I'd like to abolish. But SDL_sound looks great, thanks, I'll look into it.

jlnr:
SDL_sound seems to be in a weird limbo right now. Linux package managers have the 1.2 LGPL version, but not the [new 2.0-ish zlib-licensed version](https://www.patreon.com/posts/project-sdl-2-0-20211701).

jlnr:
The package manager status of SDL_sound doesn't even matter, as it is now an all-in-one library that I can just drop into my source tree (like utf8proc & the stb libraries). If I can get rid of the libmpg123 and libsndfile DLLs on Windows, that would be truly awesome.

The downside of SDL_sound is that I won't be able to implement Sample#length 🤣 But that is definitely a small price to pay for killing so many dependencies at once.

I'll start an SDL_sound branch of Gosu and see how well it works in practice.

jlnr:
Not yet sure if I want to rebase this branch on the latest master which uses SDL_sound.

From the SDL_sound documentation:

```cpp
/**
 * \fn Sint32 Sound_GetDuration(Sound_Sample *sample)
 * \brief Retrieve total play time of sample, in milliseconds.
 *
 * Report total time length of sample, in milliseconds. This is a fast
 *  call. Duration is calculated during Sound_NewSample*, so this is just
 *  an accessor into otherwise opaque data.
 *
 * Please note that not all formats can determine a total time, some can't
 *  be exact without fully decoding the data, and thus will estimate the
 *  duration. Many decoders will require the ability to seek in the data
 *  stream to calculate this, so even if we can tell you how long an .ogg
 *  file will be, the same data set may fail if it's, say, streamed over an
 *  HTTP connection. Plan accordingly.
 *
 * Most people won't need this function to just decode and playback, but it
 *  can be useful for informational purposes in, say, a music player's UI.
 *
 *    \param sample Sound_Sample from which to retrieve duration information.
 *   \return Sample length in milliseconds, or -1 if duration can't be
 *           determined for any reason.
 */
SNDDECLSPEC Sint32 SDLCALL Sound_GetDuration(Sound_Sample *sample);
```

hieubht-2055 (approved):

jlnr:
I'm really happy with SDL_Sound and will close this issue.

I still need to add some length limiting logic to prevent Sample.new("looping_song.s3m") from crashing (right now it tries to build an infinite buffer, but it will be more like #441.

Issue #530, opened 2020-05-07
Gosu - Failed to install Gem into Fedora 32 Silverblue


Fulgurance:
Hello, today, i try to install gosu gem into the last Fedora Silverblue, but i have this error:

```
[zohran@msi-gs73vr-6rf ~]$ gem install gosu
Building native extensions. This could take a while...
ERROR:  Error installing gosu:
	ERROR: Failed to build gem native extension.

    current directory: /var/home/zohran/.gem/ruby/gems/gosu-0.15.1/ext/gosu
/usr/bin/ruby -I /usr/share/rubygems -r ./siteconf20200507-2619-mqafhq.rb extconf.rb
The Gosu gem requires some libraries to be installed system-wide.
See the following site for a list:
https://github.com/gosu/gosu/wiki/Getting-Started-on-Linux
checking for -lopenal... yes
checking for AL/al.h... yes
creating Makefile

current directory: /var/home/zohran/.gem/ruby/gems/gosu-0.15.1/ext/gosu
make "DESTDIR=" clean
rm -f 
rm -f gosu.so  *.o  *.bak mkmf.log .*.time

current directory: /var/home/zohran/.gem/ruby/gems/gosu-0.15.1/ext/gosu
make "DESTDIR="
g++ -I. -I/usr/include -I/usr/include/ruby/backward -I/usr/include -I../../src -I../..  -I/usr/include/SDL2 -I/usr/include/AL   -I/usr/include/freetype2 -I/usr/include/libpng16 -DHAVE_AL_AL_H    -fPIC -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -std=gnu++11  -D_REENTRANT     -m64 -o Audio.o -c ../../src/Audio.cpp
Dans le fichier inclus depuis ../../src/Audio.cpp:2:
../../src/OggFile.hpp:11:32: attention: option inconnue après « #pragma GCC diagnostic »; vouliez-vous utiliser « -Wcomment » ? [-Wpragmas]
   11 | #pragma GCC diagnostic ignored "-Wcomma"
[...]

PR #532, opened 2020-05-20
Include <stdexcept> to fix compile errors on GCC 10.

jackorp:
Gem fails to compile native extensions on GCC 10. This PR fixes that by including <stdexcept> in files that need it.
I tested this by successfully building the gem, installing it, and displaying a simple window on Fedora 32.
Fixes: https://github.com/gosu/gosu/issues/530

### DIFF
--- a/src/Font.cpp
+++ b/src/Font.cpp
@@ -10,6 +10,7 @@
 #include <array>
 #include <cassert>
 #include <map>
+#include <stdexcept>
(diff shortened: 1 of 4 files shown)


jlnr:
Thanks! Sounds like this fixes #530.

jlnr:
Thanks a lot!

PR #535, opened 2020-05-24
Modernize CMake setup

jlnr:
Being able to load Gosu in CLion (through CMake) would be great for Linux development, also to get CLion's excellent refactoring & inspection support.


jlnr:
@oli-obk I'm not sure if you are still following this project. If you are, please do let me know if I'm breaking more than necessary.
My current plan is to kick Win32 support out of the CMake files, then hopefully replace all audio dependencies with a statically included SDL_sound, and finally to bring Win32 support back when SDL2 is the only dependency. I also hope that a lot of the Win32 special-casing is not necessary anymore with "modern" CMake.

oli-obk:
I was still using gosu up to last year, but not via cmake. I stopped teaching C++ last year, thus also stopped doing the gosu course. But the course still happens every year, just with someone else doing the teaching.

From my perspective, feel free to brick anything. The course has its own setup via MSVC and won't really be affected.

jlnr:
@oli-obk Great to hear that there's still a course going on 😎 I'll try to keep things sane while refactoring.

PR #539, opened 2020-07-04
Replace all audio file libraries with SDL_Sound

jlnr:
Not sure where this is going, but it definitely clashes with #527.


jlnr:
Not quite sure why the CMake build broke on macOS, but I'm going to merge this and move any CMake refactoring to a future PR.

Issue #497, opened 2019-05-02
Resizable windows start "scaling up" after a point


cyberarm:
See video: https://youtu.be/xE2LUiSjiL0

cyberarm:
Seems to be affected by https://github.com/gosu/gosu/blob/master/src/Window.cpp#L192-L193
```c++
// If the window is resizable, limit its size, without preserving the aspect ratio.
width  = actual_width  = min(width,  max_width);
height = actual_height = min(height, max_height);
```

jlnr:
Whoops, well that's not great. Thanks for reporting!

jlnr:
I have not been able to reproduce any of these issues on macOS, but things seem to be broken in various ways on Linux (elementary, based on Ubuntu 18.04 with SDL 2.0.8).

jlnr:
Specifically, if I add `resizable: true` to the gosu-examples runner, then "maximize" the window, the Tutorial appears in the wrong place, and the sidebar is drawn over other content as well. It almost looks as if the WM didn't report the window size back correctly.

jlnr:
I think one issue is that SDL_SetWindowSize does not affect maximized (not fullscreen) windows, so Gosu thinks the window size has changed when it hasn't. This might also affect Windows, not sure.

jlnr:
My current plan would be to separate the responsibilities of `Window::resize` - it should only remember some state and call SDL_SetWindowSize, and the SDL_WINDOWEVENT_SIZE_CHANGED handler should perform the necessary changes to Gosu::Graphics (OpenGL) and Gosu::Input. But I'll go and try to release a new Ruby.app for macOS first before I try something as risky as this refactoring :)

PR #541, opened 2020-08-10
Possibly fix window scaling weirdness for resizable windows

cyberarm:
Move `pimpl->resizable = resizable` to above first call to `resize`. Fixes oversized windows becoming smaller.

Removed `min(width/height, max_width/height)`. May fix scaling weirdness. #497

### DIFF
--- a/src/Resolution.cpp
+++ b/src/Resolution.cpp
@@ -28,6 +28,114 @@ int Gosu::screen_height(Window* window)
     return display_mode(window).h;
 }
 
+#ifdef GOSU_IS_MAC
+#import <AppKit/AppKit.h>
+
+static NSSize max_window_size(Gosu::Window* window)
+{
+    // The extra size that a window needs depends on its style.
+    // This logic must be kept in sync with SDL_cocoawindow.m to be 100% accurate.
+    NSUInteger style;
+    if (window && window->borderless()) {
+        style = NSWindowStyleMaskBorderless;
+    }
+    else {
+        style = (NSWindowStyleMaskTitled | NSWindowStyleMaskClosable | NSWindowStyleMaskMiniaturizable);
+    }
+    if (window && window->resizable()) {
+        style |= NSWindowStyleMaskResizable;
+    }
+
+    auto index = window ? SDL_GetWindowDisplayIndex(Gosu::shared_window()) : 0;
+    auto screen_frame = NSScreen.screens[index].visibleFrame;
+    return[NSWindow contentRectForFrameRect : screen_frame styleMask : style].size;
+}
+
+int Gosu::available_width(Window* window)
+{
+    return max_window_size(window).width;
+}
+
+int Gosu::available_height(Window* window)
+{
+    return max_window_size(window).height;
+}
+#endif
+
+// TODO: Remove this implementation and remove ifdef for GOSU_IS_X once WIN_GetWindowBordersSize is patched
+#ifdef GOSU_IS_WIN
+#include <windows.h>
+#include <SDL_syswm.h>
+#include <dwmapi.h>
+#pragma comment (lib, "Dwmapi.lib")
+
+static SDL_Rect max_window_size(Gosu::Window* window)
+{
+    // Replicate SDL's WIN_GetWindowBordersSize implementation (https://github.com/SDL-mirror/SDL/blob/e8ad67a12fff46af9927660603fdfa4cb74130be/src/video/windows/SDL_windowswindow.c#L514-L554)
+    // until it's patched to ignore the window drop shadow (window border is 1px but with drop shadow it's reported as 8px)
+    // REF: https://bugzilla.libsdl.org/show_bug.cgi?id=5302
+
+    static struct VideoSubsystem {
+        VideoSubsystem() { SDL_InitSubSystem(SDL_INIT_VIDEO); };
+        ~VideoSubsystem() { SDL_QuitSubSystem(SDL_INIT_VIDEO); };
+    } subsystem;
+
+    SDL_SysWMinfo info;
+    RECT rcClient, rcWindow;
+    POINT ptDiff;
+    SDL_Rect rect;
+    int top, left, bottom, right;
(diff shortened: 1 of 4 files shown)


jlnr:
@cyberarm Why did you close and re-push this fix? For the draft status?

cyberarm:
@jlnr Mainly to switch to source branch away from `master` (I think Github now allows setting PR as draft.)

jlnr:
The main fix here is part of gosu 1.1.0.pre1, can you please see if it fixes the issue?

cyberarm:
@jlnr the changes in 1.1.0.pre1 result in a blank window which does not seem call update and presumably draw.

jlnr:
Oh no :( Thanks for testing.

It might have to do with the failing test in #569. I'm pretty sure that gosu-examples worked just fine with 1.1.0.pre1, will have to take another look.

jlnr:
Pushed a new preview gem that should fix the issue.

cyberarm:
1.1.0.pre2 works!

cyberarm:
@jlnr I replaced the platform specific implementations of `Gosu::available_width/height` in src/Resolution.cpp with [SDL_GetDisplayUsableBounds](https://wiki.libsdl.org/SDL_GetDisplayUsableBounds), this seems to fix the weird scaling bug that this PR is meant to fix; Can you test this on Mac to ensure it doesn't break things?

cyberarm:
Just successfully tested the `available_width/height` refactor on Windows

jlnr:
Thanks, I wasn't aware of this SDL_GetDisplayUsableBounds! However, the previous implementations of available_width/available_height also did something that is maybe a bit unintuitive, they subtract the size of window decorations, which reduce the maximum width/height that a Window can have.

This can be achieved portably using SDL_GetWindowBordersSize, but only on X11 and Windows (despite the documentation saying that it's X11 only).

Can you please bring back the macOS implementation and only replace the ones for Windows/X11 for now?

cyberarm:
@jlnr I added back the Mac implementation and a Windows one due to SDL_GetWindowBordersSize reporting 8(px) when it should be 1(px). Hopefully SDL will get a patch.

jlnr:
Ah, I was wondering why you chose not to use the SDL 2 implementation on Windows. Ah well, still a great step forward for Linux :+1: Will review later.

src/Resolution.cpp
```diff
@@ -40,15 +40,15 @@ static NSSize max_window_size(Gosu::Window* window)
...
+        style = (NSWindowStyleMaskTitled | NSWindowStyleMaskClosable | NSWindowStyleMaskMiniaturizable);
     }
     if (window && window->resizable()) {
         style |= NSWindowStyleMaskResizable;
     }
 
     auto index = window ? SDL_GetWindowDisplayIndex(Gosu::shared_window()) : 0;
     auto screen_frame = NSScreen.screens[index].visibleFrame;
-    return [NSWindow contentRectForFrameRect:screen_frame styleMask:style].size;
+    return[NSWindow contentRectForFrameRect : screen_frame styleMask : style].size;
```
jlnr, line 51:
```suggestion
    return [NSWindow contentRectForFrameRect:screen_frame styleMask:style].size;
```
That's the normal way to format this in Objective-C

src/Resolution.cpp
```diff
@@ -62,67 +62,108 @@ int Gosu::available_height(Window* window)
...
-    // Because AdjustWindowRectEx will make our rect larger, not smaller, we need to perform some
-    // unintuitive math here.
-    SIZE size;
-    size.cx = 2 * (work_area.right - work_area.left) - (window_size.right - window_size.left);
-    size.cy = 2 * (work_area.bottom - work_area.top) - (window_size.bottom - window_size.top);
-    return size;
-}
+    // Replicate SDL's WIN_GetWindowBordersSize implementation (https://github.com/SDL-mirror/SDL/blob/e8ad67a12fff46af9927660603fdfa4cb74130be/src/video/windows/SDL_windowswindow.c#L514-L554)
+    // until it's patched to ignore the window drop shadow (window border is 1px but with drop shadow it's reported as 8px)
+    // REF: https://bugzilla.libsdl.org/show_bug.cgi?id=5302
```
jlnr, line 76: This fails to link because it's the first time we use a Dwm* API (see red Appveyor build). I don't have a Windows VM right now (arm64 Mac 😬), I'll try to manually add it to the MSVC project and push a commit...hope I don't break anything

src/Resolution.cpp
```diff
@@ -62,67 +62,108 @@ int Gosu::available_height(Window* window)
...
+    SDL_SysWMinfo info;
+    RECT rcClient, rcWindow;
+    POINT ptDiff;
+    SDL_Rect rect;
+    int top, left, bottom, right;
+
+    SDL_VERSION(&info.version);
+    SDL_GetWindowWMInfo(Gosu::shared_window(), &info);
+
+    HWND hwnd = info.info.win.window;
```
jlnr, line 92: This whole part of the function should go into an `if (window) {`... block

src/Resolution.cpp
```diff
@@ -40,15 +40,15 @@ static NSSize max_window_size(Gosu::Window* window)
         style = NSWindowStyleMaskBorderless;
     }
     else {
-        style = (NSWindowStyleMaskTitled|NSWindowStyleMaskClosable|NSWindowStyleMaskMiniaturizable);
+        style = (NSWindowStyleMaskTitled | NSWindowStyleMaskClosable | NSWindowStyleMaskMiniaturizable);
```
jlnr, line 43: I think I've copied the formatting from SDL_cocoawindow.m, but if we fix the spacing around | we can also drop the ()

```suggestion
        style = NSWindowStyleMaskTitled | NSWindowStyleMaskClosable | NSWindowStyleMaskMiniaturizable;
```

jlnr:
Woohoo! Not sure where Travis has gone, but this looks good to me now. :+1:

PR #542, opened 2020-08-13
Fix button ids/constants drifting by 3 for each gamepad

cyberarm:
```text
GP_BUTTON_0
GP_0_BUTTON_3
GP_1_BUTTON_6
```


jlnr (approved):
Oops 🤣 Thanks, good catch!

PR #543, opened 2020-08-13
Fix multiple gamepads at launch

cyberarm:
If multiple gamepads are present at launch (and maybe if connected simultaneously) the first gamepad is opened again instead instead of the correct gamepad.
This causes the first gamepad to be "cloned" and screws up the `gamepad_slot` system.
Since the `Joystick Instance ID` is the same for the clones which is checked when removing a gamepad, this causes gamepads to be orphaned.

Cause: `e->jdevice.which` appears to always be `0` (SDL Device ID)

This PR eagerly tries to connect to every available gamepad when a new gamepad is connected, ignoring already connected gamepads.

### DIFF
--- a/src/Input.cpp
+++ b/src/Input.cpp
@@ -159,6 +159,7 @@ struct Gosu::Input::Impl
                         );
                         gamepad_slot = available_gamepad_slot_index();
                         device_instance_id = SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(game_controller));
+                        printf("Controller connected [%i]: %s\n", e->jdevice.which, SDL_GameControllerNameForIndex(e->jdevice.which));
                     }
                 }
                 // ...but fall back on the good, old SDL_Joystick API.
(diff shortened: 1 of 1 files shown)


hieubht-2055 (approved):

src/Input.cpp
```diff
@@ -150,28 +150,40 @@ struct Gosu::Input::Impl
...
-                int device_instance_id = -1;
-                // Prefer the SDL_GameController API...
-                if (SDL_IsGameController(e->jdevice.which)) {
-                    if (SDL_GameController *game_controller = SDL_GameControllerOpen(e->jdevice.which)) {
-                        open_game_controllers.emplace_back(
-                            shared_ptr<SDL_GameController>(game_controller, SDL_GameControllerClose)
-                        );
+                int joystick_instance_id = -1;
+
+                // Loop through attached gamepads as e->jdevice.which cannot be trusted (always 0)
```
jlnr, line 155: When I start a Gosu game with two gamepads already attached on macOS, `e->jdevice.which` is 0 for the first gamepad and 1 for the second. I wonder if this is actually an SDL bug on Linux?

jlnr (approved):
Works as well as the original version for me (on macOS). I'll merge this without having tested it on Linux or Windows, but that's true of all of the new gamepad changes.

PR #544, opened 2020-08-30
CMake refactoring

jlnr:
Goals:
- [x] CI green again on macOS (cherry-picked into master)
- [x] Maybe update directory structure to move Gosu/ into include/?
- [x] Make it possible to build and run the example games straight from CLion
- [ ] Double-check what we have right now vs. popular CMake guides such as:

https://cliutils.gitlab.io/modern-cmake/chapters/basics/functions.html

https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1

https://steveire.wordpress.com/2017/11/05/embracing-modern-cmake/

https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/


jlnr:
I have now mostly followed this guide: https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/

As always with CMake, it was extremely time-consuming and quite unsatisfying. The example project has received lots of changes from PRs that I haven't had the time to understand in depth.

I've removed pkg-config support on the way because I don't use it and it's not tested through the CI. This annoys me a little, pkg-config actually is a lot more understandable than CMake.

Will probably revisit this in 2022, when Ubuntu 20.20 is the baseline.

PR #546, opened 2020-09-23
Fix missing Gosu.axis

cyberarm:
Adds missing `Gosu.axis` to Ruby/Gosu swig interface
SWIG Also picked up `Gosu::Song#length` and `Gosu::Sample#length`

### DIFF
--- a/src/RubyGosu.cxx
+++ b/src/RubyGosu.cxx
@@ -7842,37 +7842,6 @@ fail:
 
 
 SWIGINTERN VALUE
-_wrap_Sample_length(int argc, VALUE *argv, VALUE self) {
-  Gosu::Sample *arg1 = (Gosu::Sample *) 0 ;
-  void *argp1 = 0 ;
-  int res1 = 0 ;
-  double result;
-  VALUE vresult = Qnil;
-  
-  if ((argc < 0) || (argc > 0)) {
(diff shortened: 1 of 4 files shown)


src/RubyGosu.cxx
```diff
@@ -8057,6 +8095,37 @@ _wrap_Song_current_song(int argc, VALUE *argv, VALUE self) {
 }
 
 
+SWIGINTERN VALUE
+_wrap_Song_length(int argc, VALUE *argv, VALUE self) {
```
jlnr, line 8099: Wait...what? I must have messed around with all the audio branches too badly because Audio.hpp declares length() methods that are never implemented in Audio.cpp. Can you please remove them from the header files, and re-run SWIG?

jlnr (changes requested):
Thanks for catching this oversight; I'm mostly playing around with C++ recently. But you also uncovered another one on the way -> let's use this PR as a chance to fix things.

jlnr:
This also seems to be why the Appveyor build is failing.

src/RubyGosu.cxx
cyberarm: Done and done.

jlnr (approved):
Thanks! 🙏

PR #548, opened 2020-10-04
Update Visual Studio Solution

cyberarm:


Gosu/Image.hpp
```diff
@@ -8,6 +8,8 @@
 #include <Gosu/GraphicsBase.hpp>
 #include <memory>
 #include <vector>
+#include <string>
+#include <iostream>
```
jlnr, line 12:
```suggestion
#include <iostream>
#include <memory>
#include <string>
#include <vector>
```
I try to keep things alphabetical to avoid duplicates in longer lists

jlnr: But, why iostream?

windows/Gosu.vcxproj
```diff
@@ -77,14 +77,14 @@
     <IntDir Condition="'$(Configuration)|$(Platform)'=='Release|x64'">$(Platform)\$(ProjectName)-$(Configuration)\</IntDir>
     <LibraryPath Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">..\dependencies\al_soft\x86;..\dependencies\mpg123\x86;$(LibraryPath)</LibraryPath>
     <LibraryPath Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">..\dependencies\al_soft\x86;..\dependencies\mpg123\x86;$(LibraryPath)</LibraryPath>
-    <LibraryPath Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">..\dependencies\al_soft\x64;..\dependencies\mpg123\x64;$(LibraryPath)</LibraryPath>
-    <LibraryPath Condition="'$(Configuration)|$(Platform)'=='Release|x64'">..\dependencies\al_soft\x64;..\dependencies\mpg123\x64;$(LibraryPath)</LibraryPath>
+    <LibraryPath Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">C:\Users\cyber\Code\gosu\dependencies\SDL2\lib\x64;..\dependencies\al_soft\x64;..\dependencies\mpg123\x64;$(LibraryPath)</LibraryPath>
+    <LibraryPath Condition="'$(Configuration)|$(Platform)'=='Release|x64'">C:\Users\cyber\Code\gosu\dependencies\SDL2\lib\x64;..\dependencies\al_soft\x64;..\dependencies\mpg123\x64;$(LibraryPath)</LibraryPath>
```
jlnr, line 81: This shouldn't have absolute paths in it

windows/Gosu.vcxproj
```diff
@@ -287,23 +291,54 @@
...
     <ClInclude Include="..\src\stb_image_write.h" />
+    <ClInclude Include="..\src\TexChunk.hpp" />
+    <ClInclude Include="..\src\TextBuilder.hpp" />
+    <ClInclude Include="..\src\Texture.hpp" />
+    <ClInclude Include="..\src\TransformStack.hpp" />
+    <ClInclude Include="..\src\TrueTypeFont.hpp" />
     <ClInclude Include="..\src\WinUtility.hpp" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="..\src\GosuAppDelegate.mm" />
```
jlnr, line 340: This one is not needed on Windows

jlnr (commented):
Thanks! I have been meaning to update to a more recent MVSC version but got stuck installing all the audio dependencies. Makes sense to update now. I just need to meditate a bit on how to continue my Windows 10 development, my VM running under macOS is slow as molasses.

Gosu/Image.hpp
cyberarm: Removed iostream, I thought it was, for some odd reason, required to fix a `missing type specifier - int assumed.` error but after cleaning and rebuilding with out it, it works... 🤷🏼

cyberarm:
@jlnr AppVeyor works all the way through tests but segfaults when done... Would [properly updating](https://github.com/gosu/gosu/blob/master/dependencies/msvcrt-ruby/README.md) the Ruby `config.h` files possibly fix it?

jlnr:
I've added the rest of the SDL 2 files for MSVC to keep development on Windows nice and easy. We've already bundled the DLLs for Windows, anyway.

The Ruby error probably stems from the fact that I've used the MSVC 2015 command line tools to configure every single version of Ruby when adding the headers and libraries. So uh...I guess I could do it again for all Ruby versions (twice for 32/64) or we just remove the #error line and see if it works anyway. I've done it before (when Ruby itself still shipped header files for MSVC), and nothing exploded, it's just really ugly.

jlnr:
Support for Ruby 2.4 and earlier has ended anyway, probably a good chance to require Ruby 2.6+ (maybe 2.5+)?

jlnr:
I just realized that my version of MSVC 2019, after the latest update, complained because _MSC_VER was not the version you required. So it seems the _MSC_VER is granular enough that it even changes across MSVC 2019 versions. I've removed all the preprocessor checks while I'm investigating the segfault.

PR #550, opened 2020-11-15
Fix AppVeyor build

jlnr:
Not sure where the segfault comes from, the current master branch works really well with both Ruby 2.6 and 2.7 for me on Windows 10.

Running the tests also doesn't crash. This is a low-priority PR because having a CI would be nice, but isn't a blocker for 1.0.


jlnr:
Okay, narrowed it down to the window tests. Probably unrelated, but when I ran `rake tests_interactive`, something about the drag & drop test wrong...I think the window just disappeared?

jlnr:
If I attach a debugger to the Ruby process before it crashes, I can see that the SIGSEGV comes from the onexit/atexit table in gosu.so. Let's see what happens if I disable all of Gosu's atexit blocks.

jlnr:
Let's see if AppVeyor works again if I remove all the std::atexit / Ruby at_exit magic from Gosu. Freeing resources before the process exits is practically pointless, and whatever I did right now hasn't gotten rid of the following warning:

> AL lib: (EE) alc_cleanup: 1 device not closed

Instead of understanding the lifecycle of Windows DLLs and Ruby C extensions...let's just try to sidestep the issue.

PR #552, opened 2020-11-29
Bitmap cleanup

jlnr:
Now that Gosu can be loaded and refactored in CLion, cleaning up the C++ interface for Gosu 1.0 has become much easier.

* [x] Add .clang-tidy, .clang-format files (not perfect yet)
* [x] Add Google Test/ctest unit tests
* [x] Clean up Bitmap.{hpp,cpp} and fixes warnings (had to touch a few dependent files as well)
* [x] Clean up BitmapIO.cpp
* [x] Add a unit test that uses one of the media files


Kjarrigan (approved):
LGTM I only have to "notes:

* The indentation looks a bit off in some files. For example: examples/Tutorial/CMakeLists.txt  In the diff it's not really visible but if you look at the file two of the multiline blocks look wrong
* `Bitmap.insert` Not too sure if swapping the args is necessary. I see that it might be more consistent with stuff like resize but overall bitmap dot insert for me sounds like I have to provide a bitmap and were to insert it. But I do not use the C interface so I don't mind the swap.

jlnr:
@Kjarrigan Thanks! Good catch about the tab characters in the CMakeList files. Will clean that up.

About Bitmap::insert, I'm a bit torn myself. The Ruby interface is `Image#insert(img, x, y)`, and it feels dumb to reorder Bitmap::insert the other way around. However, I also think that it makes sense to have the `src_*` arguments right after the source 🤔

src/BitmapIO.cpp
```diff
@@ -21,73 +19,67 @@
...
     callbacks.skip = skip_callback;
     callbacks.eof = eof_callback;
-    int x, y, n;
-    stbi_uc* bytes = stbi_load_from_callbacks(&callbacks, &input, &x, &y, &n, STBI_rgb_alpha);
+    int x = 0, y = 0, n = 0;
 
+    stbi_uc* bytes = stbi_load_from_callbacks(&callbacks, &input, &x, &y, &n, STBI_rgb_alpha);
     if (bytes == nullptr) {
-        // TODO - stbi_failure_reason is not thread safe. Everything here should be wrapped in a
-        // mutex.
```
jlnr, line 79: Recent versions of stb_image use a thread-local stbi_failure_reason \o/

test/BitmapTests.cpp
```diff
@@ -0,0 +1,96 @@
+#include <gtest/gtest.h>
+#include <Gosu/Bitmap.hpp>
+
+namespace Gosu
+{
+    class BitmapTests : public testing::Test
+    {
+    };
+
+    static testing::AssertionResult visible_pixels_are_equal(const Bitmap& lhs, const Bitmap& rhs)
```
jlnr, line 10: I'm not really happy with the crazy indentation of these tests. Maybe I'll reduce the namespace indent from 4 to 0, after all.

PR #555, opened 2020-12-08
Add support for cursor capture and relative mouse motion

cyberarm:
* Mouse capture via [SDL_SetWindowGrab](https://wiki.libsdl.org/SDL_SetWindowGrab)
* Relative mouse motion via [SDL_SetRelativeMouseMode](https://wiki.libsdl.org/SDL_SetRelativeMouseMode)

Useful for windowed mouse look games like jcaster, i-mic-fps, etc.

### DIFF
--- a/include/Gosu/Window.hpp
+++ b/include/Gosu/Window.hpp
@@ -89,6 +89,9 @@ namespace Gosu
         //! If this function returns true, the system cursor will be visible while over the window.
         virtual bool needs_cursor() const { return true; }
 
+        //! If this function returns true, the system cursor is confined to the window.
+        virtual bool capture_cursor() const { return false; }
+
         virtual void gain_focus() {}
 
         //! This function is called when the window loses focus on some platforms.
--- a/include/Gosu/Window.hpp
+++ b/include/Gosu/Window.hpp
@@ -92,6 +92,11 @@ namespace Gosu
         //! If this function returns true, the system cursor is confined to the window.
         virtual bool capture_cursor() const { return false; }
 
+        //! If this function returns true, the system cursor is hidden and the mouse is unconfined by the display size.
+        //! Note: Gosu::mouse_x and Gosu::mouse_y will stop updating if this is set to true, query Gosu::relative_mouse_x
+        //! and Gosu::relative_mouse_y instead.
+        virtual bool relative_mouse_mode() const { return false; }
+
         virtual void gain_focus() {}
 
         //! This function is called when the window loses focus on some platforms.
--- a/ext/gosu/gosu.i
+++ b/ext/gosu/gosu.i
@@ -1083,6 +1083,7 @@ namespace Gosu
 %rename("mouse_x=") set_mouse_x;
 %rename("mouse_y=") set_mouse_y;
 %rename("needs_cursor?") needs_cursor;
+%rename("capture_cursor?") capture_cursor;
 %rename("needs_redraw?") needs_redraw;
 %rename("close!") force_close;
 %rename("fullscreen?") fullscreen;
--- a/ffi/Gosu_FFI_internal.h
+++ b/ffi/Gosu_FFI_internal.h
@@ -130,6 +130,12 @@ struct Gosu_Window : public Gosu::Window
         return Window::needs_cursor();
     }
 
+   bool capture_cursor() const override
+    {
+        if (capture_cursor_callback) return capture_cursor_callback();
+        return Window::capture_cursor();
+    }
+
     void close() override
     {
         if (close_callback != nullptr) return close_callback();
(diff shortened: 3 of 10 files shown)


include/Gosu/Window.hpp
```diff
@@ -34,7 +34,8 @@ namespace Gosu
         //! \param update_interval Interval in milliseconds between two calls to the update member
         //! function.
         Window(unsigned width, unsigned height, bool fullscreen = false,
-            double update_interval = 16.666666, bool resizable = false);
+            double update_interval = 16.666666, bool resizable = false,
+            bool borderless = false);
```
jlnr, line 38: At this point I think we should merge fullscreen, resizable, and borderless into a single `unsigned flags` and introduce a `WindowFlags` enum akin to the existing `ImageFlags`.

jlnr: Gosu::WF_FULLSCREEN should have the value 1 for legacy code that passes `true` for this `unsigned` argument.

cyberarm:
@jlnr What are your thoughts on this PR?
* The hit test callback will enable borderless windows to be draggable and resizable
* The capture_cursor callback will enable applications that rely on the mouse being confined to the window area (e.g. FPS camera controller without jumping from lost input because mouse left window)

jlnr:
@cyberarm Mixed:

* I love the idea of capturing the cursor.
* After using Macs for 15 years, having a "maximize" function seems very Windows-like, and I'd have to check how it even works on macOS. (Usually apps just use fullscreen mode.) I'm also not sure when an app that I used ever minimized or restored itself. What are the use cases for it (sorry if I've asked this before)?
* Similarly skeptical about the hit-testing use cases, and having another enum in Gosu is not something I am looking forward to :/

I guess the dividing line is that I'd merge anything that helps building great 2D games for desktop and mobile, but I need some convincing with almost-finished projects for anything that goes into the UI toolkit area.

cyberarm:
> * After using Macs for 15 years, having a "maximize" function seems very Windows-like, and I'd have to check how it even works on macOS. (Usually apps just use fullscreen mode.) I'm also not sure when an app that I used ever minimized or restored itself. What are the use cases for it (sorry if I've asked this before)?
> * Similarly skeptical about the hit-testing use cases, and having another enum in Gosu is not something I am looking forward to :/

Having minimize, maximize and restore functions would be useful for windowed borderless applications and games that want a fully custom appearance, e.g. the [TimeCrafters Configuration Tool](https://github.com/TimeCrafters/timecrafters_configuration_tool_desktop) that I wrote for the robotics team would benefit from both the hit-testing and min/maximize functions.
Screenshot-2021-07-19-080332

jlnr:
Aah, I see! Makes sense that borderless windows would need a way to recreate the window decorations. I have to say I prefer how some portable apps (Spotify comes to mind) have native macOS window controls in a non-native window. I've never looked into how it works, but that seems much nicer than manually recreating controls that don't respect the rest of the OS (left vs right window edge, some distros don't have a minimize button, etc.)

Bildschirmfoto 2021-08-21 um 12 29 35

(It doesn't seem to be trivial because Spotify doesn't do the same thing on Linux :/ It'd probably have to be in SDL2 proper for me to consider it)

Anyway, that's pretty deep in UI toolkit land for me. I'd prefer to only merge the cursor capture for now.

cyberarm:
@jlnr I've removed the window controls and hit testing from this PR.
I think I added everything needed to merge except for a git diff patch for `capture_cursor?` ➡️ `protected_capture_cursor?` in `gosu_SWIG_RENAME_PATCH.patch`

jlnr (commented):
1. Can you please rebase on master?
2. Hmm, isn't this PR missing something to work for something like a 3D shooter? Because if you were to implement mouse camera movements, you wouldn't just want the cursor to be confined to the window, because then once you hit any edge, you cannot move any further. Probably not an issue in other frameworks that have a mouseMove(dx, dy) event, but Gosu doesn't. :/ Or is the intention that users of capture_mouse still need to move the mouse back to the window center every tick?

cyberarm:
>     Or is the intention that users of capture_mouse still need to move the mouse back to the window center every tick?
This was the original intention, however we could use [SDL_SetRelativeMouseMode](https://wiki.libsdl.org/SDL_SetRelativeMouseMode) and add `Window#relative_mouse_x/y` to accomplish the same with perhaps more consistent/smoother mouse movement since the mouse doesn't need to be re-centered each `update`. 

`SDL_SetRelativeMouseMode` does hide the cursor and prevent it from "moving" according to the wiki, so it might be desirable to keep `capture_cursor` and add another callback, perhaps `Window#relative_mouse_motion` to enable/disable relative mouse motion.

PR #556, opened 2020-12-09
Compatibility fixes

jlnr:
* Makes some Image#draw arguments optional again, for compatibility with older games
* Don't set CMP0072, which was only introduced in CMake 3.11 (Gosu only requires 3.10); instead, confirm that we agree with all policy changes inbetween CMake 3.10 and 3.18.4: https://cmake.org/cmake/help/git-stage/command/cmake_policy.html#setting-policies-by-cmake-version

### DIFF
--- a/rdoc/gosu.rb
+++ b/rdoc/gosu.rb
@@ -475,7 +475,7 @@ module Gosu
     # @see #draw_as_quad
     # @see https://github.com/gosu/gosu/wiki/Basic-Concepts#drawing-with-colours Drawing with colors, explained in the Gosu Wiki
     # @see https://github.com/gosu/gosu/wiki/Basic-Concepts#z-ordering Z-ordering explained in the Gosu Wiki
-    def draw(x, y, z, scale_x=1, scale_y=1, color=0xff_ffffff, mode=:default); end
+    def draw(x, y, z=0, scale_x=1, scale_y=1, color=0xff_ffffff, mode=:default); end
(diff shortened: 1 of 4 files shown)


rdoc/gosu.rb
```diff
@@ -475,7 +475,7 @@ def subimage(left, top, width, height); end
     # @see #draw_as_quad
     # @see https://github.com/gosu/gosu/wiki/Basic-Concepts#drawing-with-colours Drawing with colors, explained in the Gosu Wiki
     # @see https://github.com/gosu/gosu/wiki/Basic-Concepts#z-ordering Z-ordering explained in the Gosu Wiki
-    def draw(x, y, z, scale_x=1, scale_y=1, color=0xff_ffffff, mode=:default); end
+    def draw(x, y, z?ß, scale_x=1, scale_y=1, color=0xff_ffffff, mode=:default); end
```
cyberarm, line 478: `z?ß` shouldn't that be `z=0`?

jlnr: Whoops, patched the docs as the last thing before I went to bed. Thanks for catching it, chances are the CI wouldn't have! :+1:

jlnr:
Not sure why I'm getting random failures now where Travis cannot find a display on Linux. However, I also don't feel like investing more in the Travis setup, that time is probably better spent on migrating to GitHub actions sooner or later.

PR #561, opened 2020-12-27
Ruby 3.0 support

jlnr:
FYI @mblumtritt @chesterbr this fixes the issue for me.

The bigger challenge will be to come up with a release plan for this, given that the master branch is not super stable yet 🤔

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,9 @@ matrix:
       dist: bionic
       sudo: required
       compiler: gcc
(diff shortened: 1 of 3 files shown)


jlnr:
Hmm, now the macOS 10.14 build fails because `sdl2-config --static-libs` contains `-Wl,-weak_framework,CoreHaptics`, but CoreHaptics.framework does not exist on the build machine. Sounds like a bug in either SDL 2 or Homebrew, but maybe I should hotfix it as well.

Issue #560, opened 2020-12-25
Unable to compile Gosu with latest Ruby version


mblumtritt:
I sadly found today that Gosu is unable to compile with latest Ruby version 3.0.0. 😢

How can I help to fix this?

chesterbr:
Indeed, it seems to be using `rb_check_safe_obj`, which seems to have been deprecated in 2.7 and removed in 3.0:

```
 ../../src/RubyGosu.cxx:2281:9: error: use of undeclared identifier 'rb_check_safe_obj'; did you mean 'rb_check_safe_str'?
```

<details>
<summary>full build log</summary>

```
current directory: /Users/chesterbr/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/gosu-0.15.2/ext/gosu
/Users/chesterbr/.rbenv/versions/3.0.0/bin/ruby -I /Users/chesterbr/.rbenv/versions/3.0.0/lib/ruby/3.0.0 -r ./siteconf20201226-56320-ny1ega.rb extconf.rb
The Gosu gem requires some libraries to be installed system-wide.
See the following site for a list:
https://github.com/gosu/gosu/wiki/Getting-Started-on-OS-X
creating Makefile

current directory: /Users/chesterbr/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/gosu-0.15.2/ext/gosu
make "DESTDIR=" clean

current directory: /Users/chesterbr/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/gosu-0.15.2/ext/gosu
make "DESTDIR="
compiling ../../src/Audio.cpp
../../src/Audio.cpp:51:9: warning: 'alGenBuffers' is deprecated: first deprecated in macOS 10.15 - OpenAL is deprecated in favor of AVAudioEngine [-Wdeprecated-declarations]
        alGenBuffers(1, &buffer);
        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/OpenAL.framework/Headers/al.h:592:25: note: 'alGenBuffers' has been explicitly marked deprecated here
AL_API void AL_APIENTRY alGenBuffers( ALsizei n, ALuint* buffers ) OPENAL_DEPRECATED;
                        ^
[...]

PR #562, opened 2020-12-27
Merge Song & Sample (Audio 3.0, supersedes #441)

jlnr:
One "issue" with using SDL_sound for audio loading is that it supports formats with built-in loop points. When Gosu::Sample#initialize tries to completely preload such a file, it will crash.

This seems like a great chance to merge Sample & Song by having them both stream the file by default. If it is short enough, it can stay loaded in memory as an optimization (= the old Sample behavior).

This was meant to land in Gosu 1.0, but given #560, I think I will push this back since it is easy enough to work around this issue during development by not using looping samples.

### DIFF
--- a/examples/Tutorial-Touch/Tutorial-Touch.cpp
+++ b/examples/Tutorial-Touch/Tutorial-Touch.cpp
@@ -62,7 +62,7 @@ public:
 class Player
 {
     Gosu::Image image;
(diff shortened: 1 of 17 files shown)


jlnr:
Theoretically I'd still love to do this, but I don't think Gosu has time for new features right now, especially ones that invalidate existing documentation.

PR #567, opened 2021-01-01
Extract FFI sources and headers into their own directory

jlnr:
Decided to go with "ffi" for the folder name after all, because the term seems to be quite general and not linked to libffi.

I've also renamed the files to match their "class name" to distinguish them better from the C++ interface. The FFI library is now installed as `libgosu-ffi` and I can safely turn `libgosu` into a static library (in a later PR).

Will try to push the Windows change for the separate DLL now too.

To Do:

- [x] Test if the FFI DLL works (if possible?)
- [x] Update `rake win:vcxproj` to also update headers and the FFI project
- [x] Go over FFI files and review CLion/clang-tidy warnings


jlnr:
@cyberarm I've added an MSVC project for the FFI DLL, which is built either as `lib\gosu-ffi.dll` or `lib64\gosu-ffi.dll`. Do you have a way to test if this works with Crystal etc. right now?

cyberarm:
The MSVC project successfully builds, however the functions don't have the required `__declspec(dllexport)`/`__declspec(dllimport)` so the list of exported functions is just from utf8proc:

```text
C:\Users\cyber\Code\gosu\lib64>dumpbin /exports gosu-ffi.dll
Dump of file gosu-ffi.dll

File Type: DLL

  Section contains the following exports for gosu-ffi.dll

   ...

    ordinal hint RVA      name

          1    0 00037200 utf8proc_NFC = utf8proc_NFC
          ...

C:\Users\cyber\Code\gosu\lib64>
```

@jlnr Looks like we need to add this and prefix each function with with it.
It also may be necessarily to move Gosu_Constants.cpp to a header file.
I can work on this later this weekend and push to this branch if convenient?
```cpp
/* Add check for non-msvc build env */
#ifdef GOSUFFI_EXPORTS
#define GOSU_FFI_API __declspec(dllexport)
#else
#define GOSU_FFI_API __declspec(dllimport)
#endif
```

cyberarm:
@jlnr I made the aforementioned changes and pushed them to my fork: https://github.com/cyberarm/gosu/tree/extract-ffi (https://github.com/cyberarm/gosu/commit/7800a9e6c58f957c36ca9a8e1cac419096133d91)

The `ffi_helper.h` will need more work but everything else should be good. 🤞🏼

cyberarm:
The `ffi_helper.h` can probably be simplified a bit by using utf8proc's method: https://github.com/gosu/gosu/blob/30339c5e0f658f2a273a2609a27a6593ce7c8330/dependencies/utf8proc/utf8proc.h#L122-L138

jlnr:
I've reorganized the commits a bit for rebasing (didn't want to squash everything into one commit).

@cyberarm I've also removed the Gosu_Window_is_button_down function, which duplicates Gosu_button_down. Ruby/Gosu still has both variants (module and window method) because the scoping inside the window is convenient and changing that would break too much code. In my opinion this is up to the language wrappers to decide.

PR #568, opened 2021-01-17
FFI cleanup

jlnr:
Sorry for messing so much with your code @cyberarm, I wanted to play around with clang-tidy & clang-format mainly because I've been configuring the two for my current client project as well. I've gotten rid of a lot of warnings and reinterpret_casts mostly to see if I can satisfy the tooling without making the code completely impenetrable.

_I think_ most of the changes are neutral or good, but the auto-aligned argument lists are definitely not great. All the FFI files will need at least one more iteration to add error reporting, I'll try to manually tweak for readability again when that happens.


.clang-tidy
```diff
@@ -5,6 +5,7 @@ Checks: >-
   -cppcoreguidelines-avoid-magic-numbers,
   -cppcoreguidelines-avoid-c-arrays,
   -cppcoreguidelines-init-variables,
+  -cppcoreguidelines-pro-bounds-pointer-arithmetic,
```
jlnr, line 8: This rule doesn't even allow `pointer++` :(

ffi/Gosu.cpp
```diff
@@ -1,251 +1,200 @@
...
 {
-    std::function<void ()> callback;
-    callback = [=]() { function(data); };
-    return reinterpret_cast<Gosu_Image *>(new Gosu::Image(Gosu::Graphics::record(width, height, callback)));
+    Gosu::Image image = Gosu::Graphics::record(width, height, [=] { function(data); });
+    return new Gosu_Image{image};
 }
 
-void Gosu_flush()
+GOSU_FFI_API void Gosu_flush(void)
```
jlnr, line 26: @cyberarm `Gosu_flush()` declares a vararg function in C (same as `Gosu_flush(...)`), but we explicitly want zero arguments. Hm, technically I could have left out the `void` here because this is C++, but it needs to be there in the C headers.

ffi/Gosu.cpp
```diff
@@ -1,251 +1,200 @@
...
+GOSU_FFI_API const char* Gosu_language()
 {
     static thread_local std::string language;
     language = Gosu::language();
-
     return language.c_str();
 }
 
-long Gosu_milliseconds()
+GOSU_FFI_API uint64_t Gosu_milliseconds()
```
jlnr, line 190: I've used types with a specific size where it matters (here and in Gosu_Color.hxx); would it make sense to make the size explicit _everywhere_? 🤔

ffi/Gosu_Image.cpp
```diff
@@ -1,168 +1,168 @@
...
+GOSU_FFI_API Gosu_GLTexInfo* Gosu_Image_gl_tex_info_create(Gosu_Image* image)
 {
-    Gosu::Image *gosu_image = reinterpret_cast<Gosu::Image *>(image);
-    Gosu::GLTexInfo *gosu_texture_info = new Gosu::GLTexInfo(*gosu_image->data().gl_tex_info());
-    if (gosu_texture_info) {
-        return reinterpret_cast<Gosu_GLTexInfo *>(gosu_texture_info);
-    }
-    else {
-        return nullptr;
-    }
```
jlnr, line 100: Casting nullptr directly doesn't hurt, it will still be nullptr.

ffi/Gosu_Image.cpp
```diff
@@ -1,168 +1,168 @@
...
 
-unsigned Gosu_Image_height(Gosu_Image *image)
+GOSU_FFI_API unsigned Gosu_Image_height(Gosu_Image* image)
 {
-    return reinterpret_cast<Gosu::Image *>(image)->height();
+    return image->image.height();
 }
 
-Gosu_GLTexInfo *Gosu_Image_gl_tex_info_create(Gosu_Image *image)
+GOSU_FFI_API Gosu_GLTexInfo* Gosu_Image_gl_tex_info_create(Gosu_Image* image)
```
jlnr, line 109: At this point I really regret that C++ Gosu uses namespace at all, imagine how easy things would be if the naming convention was just GosuGLTexInfo, GosuImage, GosuWindow and both languages could share the same definition...

ffi/Gosu_Song.cpp
```diff
@@ -1,52 +1,51 @@
...
+}
+
+GOSU_FFI_API void Gosu_Song_set_volume(Gosu_Song* song, double volume)
+{
+    return song->set_volume(volume);
+}
+
+GOSU_FFI_API Gosu_Song* Gosu_Song_current_song(void)
+{
     return reinterpret_cast<Gosu_Song*>(Gosu::Song::current_song());
```
jlnr, line 50: This is the one reinterpret_cast that I couldn't get rid of. dynamic_cast doesn't work because Song is not polymorphic.

Probably not worth worrying about, because the whole concept of Gosu::Song::current_song will hopefully be obsolete sooner rather than later.

PR #572, opened 2021-01-24
Fixes for Apple Silicon

jlnr:


jlnr:
It still doesn't work... 🤔

> /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:130:in `require': dlopen(/Users/jlnr/.gem/my_gems/gems/gosu-1.1.1/lib/gosu.bundle, 0x0009): missing compatible arch in /Users/jlnr/.gem/my_gems/gems/gosu-1.1.1/lib/gosu.bundle - /Users/jlnr/.gem/my_gems/gems/gosu-1.1.1/lib/gosu.bundle (LoadError)
> % lipo -info /Users/jlnr/.gem/my_gems/gems/gosu-1.1.1/lib/gosu.bundle
> Architectures in the fat file: /Users/jlnr/.gem/my_gems/gems/gosu-1.1.1/lib/gosu.bundle are: x86_64 arm64e 
> % arch
> arm64


So...I am on an arm64 system, the Gosu gem is compiled for Intel and "arm64e" but there's no compatible architecture? Is the system Ruby in Big Sur broken and can't decide whether arm64 or arm64e is the right identifier? What a mess :|

jlnr:
Okay, it works with Ruby 2.7 from homebrew. Not sure why it didn't work the first time I tried? Anyway, merging this without further regard for Apple's deprecated system Ruby.

PR #573, opened 2021-01-28
More robust solution of linear system in Macro::find_transform_for_target (fixes #571)

### LINKED ISSUE #571: Macro::draw uses wrong transformation when x2 == x4
Drawing a Macro as a quad computes a projective transform that maps the Macro Image's corners to the given coordinates (`x1`, `y1`), (`x2`, `y2`), (`x3`, `y3`), (`x4`, `y4`).

There seems to be a bug if `x2 == x4`, i.e. if the right edge of the quad is exactly vertical.

Here's a reference image of an Image drawn in a trapezoid shape with `x2 == x4`:

trapezoid_two_triangles

Converting the Image to a Macro and drawing the same shape yields the following wrong result (note how the computed transformation no longer maps the corners to the correct shape):

trapezoid_as_macro

If I slightly perturb the corner positions such that `x2 != x4`, it snaps back to the expected result:

trapezoid_as_macro_perturbed

jsb:
Instead of solving the lower right 2x2 subsystem iteratively using Gaussian Elimination, we can use the closed-form 2x2 matrix inverse to solve this part directly.

This avoids the branching based on certain cell values, which probably contains a bug: #571. This bug is fixed with this update.

### DIFF
--- a/src/Macro.cpp
+++ b/src/Macro.cpp
@@ -1,40 +1,30 @@
 #include "Macro.hpp"
 #include "DrawOpQueue.hpp"
 #include <Gosu/Image.hpp>
-#include <cmath>
-#include <algorithm>
-#include <functional>
-#include <memory>
 #include <stdexcept>
 using namespace std;
-
-typedef double Float;
-
-namespace
-{
-
-// Solves the 2x2 linear system for x:
-// (a11 a12) (x1) = (b1)
-// (a21 a22) (x2) = (b2)
-// x1, x2 are output parameters. Returns false if the matrix is singular.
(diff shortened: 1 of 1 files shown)


jlnr:
Thanks so much for the patch! I've verified that everything looks good with the gosu-quad-test project :+1:

Dragging these vertices around, I've also wondered whether there is any good solution to handling non-convex quads...but I honestly can't come up with a good solution right now. At least it doesn't crash :)

I've moved the helper out of the anonymous namespace. Gosu tends to use static functions instead, but here it seemed a bit more consistent to just move it into the Impl struct. I've also removed the `Float` typedef. I think I had it because the original code used to be a template.

PR #575, opened 2021-02-07
Add Android support

cyberarm:
Draft PR for tracking changes needed to compile for Android

See: https://github.com/cyberarm/gosu-android for Android project
Current goal is to get a [very basic gosu app](https://github.com/cyberarm/gosu-android/blob/3c4248eae3f18086e00b5c3b9bfd7d3192ec3be9/app/jni/src/gosu_app.cpp) running

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,0 +1,66 @@
+LOCAL_PATH := $(call my-dir)
+
+###########################
+#
+# Gosu shared library
+#
+###########################
+
+include $(CLEAR_VARS)
+
+LOCAL_MODULE := gosu
+
+LOCAL_C_INCLUDES := $(LOCAL_PATH)/include \
+	$(wildcard $(LOCAL_PATH)/src/*.hpp)
+
+LOCAL_EXPORT_C_INCLUDES := $(LOCAL_C_INCLUDES)
+
+LOCAL_SRC_FILES := \
+	$(wildcard $(LOCAL_PATH)/src/*.cpp)
+#	$(LOCAL_PATH)/src/Audio.cpp \
+#	$(LOCAL_PATH)/src/AudioFile.cpp \
+#	$(LOCAL_PATH)/src/AudioFileSDLSound.cpp \
+#	$(LOCAL_PATH)/src/AudioImpl.cpp \
+#	$(LOCAL_PATH)/src/AudioImpl.hpp \
+#	$(LOCAL_PATH)/src/Bitmap.cpp \
+#	$(LOCAL_PATH)/src/BitmapIO.cpp \
+#	$(LOCAL_PATH)/src/BlockAllocator.cpp \
+#	$(LOCAL_PATH)/src/Channel.cpp \
+#	$(LOCAL_PATH)/src/ClipRectStack.hpp \
+#	$(LOCAL_PATH)/src/Color.cpp \
+#	$(LOCAL_PATH)/src/DrawOp.hpp \
+#	$(LOCAL_PATH)/src/DrawOpQueue.hpp \
+#	$(LOCAL_PATH)/src/EmptyImageData.hpp \
+#	$(LOCAL_PATH)/src/Font.cpp \
+#	$(LOCAL_PATH)/src/Graphics.cpp \
+#	$(LOCAL_PATH)/src/Iconv.cpp \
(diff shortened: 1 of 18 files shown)


cyberarm:
@jlnr Do you know why the app crashes when `FileAndroid::read` returns?
The expected log output of `"Read asset."` from FileAndroid is printed but the log in IO of `"Loaded file: filename"` doesn't.

Call tree:
```cpp
Gosu::Image("gosu.png");
```
https://github.com/gosu/gosu/blob/0a58b3947b7288693a6686407c674741f95ed323/src/BitmapIO.cpp#L55-L62
https://github.com/gosu/gosu/blob/0a58b3947b7288693a6686407c674741f95ed323/src/IO.cpp#L52-L59
https://github.com/gosu/gosu/blob/0a58b3947b7288693a6686407c674741f95ed323/src/FileAndroid.cpp#L72-L80

```text
2021-02-08 14:45:17.065 8048-8103/org.libsdl.app I/Gosu: Loading image: gosu.png
...
(List of files in assets/ dir)
2021-02-08 14:45:17.066 8048-8103/org.libsdl.app I/Gosu: Asset filename: gosu.png
...
2021-02-08 14:45:17.066 8048-8103/org.libsdl.app I/Gosu: Opening asset: gosu.png
2021-02-08 14:45:17.066 8048-8103/org.libsdl.app I/Gosu: Reading file: gosu.png
2021-02-08 14:45:17.066 8048-8103/org.libsdl.app I/Gosu: Reading asset with length: 1617/1617...
2021-02-08 14:45:17.067 8048-8103/org.libsdl.app I/Gosu: Read asset.
2021-02-08 14:45:17.067 8048-8103/org.libsdl.app A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30518635e7854 in tid 8103 (SDLThread), pid 8048 (SDLActivity)
```

jlnr:
@cyberarm You are passing `&dest_buffer ` into AAsset_read, which will read into the memory that starts with the `dest_buffer` parameter on the stack. The extra & is the problem.

cyberarm:
Pointer, reference, value... 🤦🏽‍♂️ Thanks, @jlnr, it compiles now. 😀 
photo5557551483772315565

cyberarm:
@jlnr How do we want to implement Gosu's touch callbacks?
SDL provides us with 3 touch events `SDL_FINGERMOTION`, `SDL_FINGERDOWN`, and `SDL_FINGERUP` [SDL TouchFingerEvent](https://wiki.libsdl.org/SDL_TouchFingerEvent)

SDL doesn't provide a `touch_cancelled` event.

Do we want to keep the struct based callback?
```cpp
struct Touch {
  int id; // device_id + finger_id?
  int device_index, finger_index;
  double x, y;
};

void touch_began(Touch touch);
void touch_moved(Touch touch);
void touch_ended(Touch touch);
```
or use a more verbose callback?
```cpp
void touch_start(int device_index, int finger_index, double x, double y);
void touch_move(int device_index, int finger_index, double x, double y);
void touch_end(int device_index, int finger_index, double x, double y);
```

jlnr:
Hmm, good question. Some thoughts (after a busy couple of weeks):

* I like the struct because it's easier to add stuff like the pressure later on.
* I also like the concept of touch_cancelled. I think for the most part it can be replaced with the lose_focus callback – basically, if something happens in the foreground while the finger is down, we should not count the lost (cancelled) touch as a button tap, and so on.
  Let's leave it for now, even if only iOS calls it (which is already the case for the lose_focus callback...).
* Yes, I think we should try to make it a single `id`. Maybe hashing both 'id's together is safer than relying on adding or anything else, but we can always tweak it later.

cyberarm:
@jlnr what is the best way for `Gosu::Input` to access `Gosu::Window`'s instance of `Gosu::Graphics` for `width()` and `height()` to convert SDL's normalized (0..1) screen touch coordinates to window coordinates?

jlnr:
@jlnr It is actually supposed to work the other way around: Window should set the mouse factors of its Input object to scale correctly. Technically touches are not a mouse, but can we reuse the same mechanism? (Sorry, still no time to dig into the code a lot...)

PR #583, opened 2021-02-28
Fix macOS font styles, use SDL_locale.h for Gosu.language

jlnr:
@cyberarm I wanted to propose some FFI error handling stuff, but had two tests fail either locally or on the CI. This branch intends to fix both of them:

* Using bold and/or italic fonts on macOS was broken (at least on Big Sur?)
* Gosu.language returned `C.UTF8` or something like that, causing a test in `test_gosu_module.rb` to fail

I've (hopefully?) fixed the latter issue by throwing away all my platform-specific code in favor of using `SDL_GetPreferredLocales`.

I'm not yet 100% sure if the new `Gosu.languages` method is really the best approach. Typically what you want is to find the first language that the user understands _and_ the game supports. But then why doesn't Gosu directly perform this logic for you, e.g. you call `Gosu.language(supported_languages: ['en', 'fr'])` and it returns the first of the user's language code that starts with en or fr? 🤔 Bonus advantage: No new method because `Gosu.language()` without any arguments could just default to the previous behavior.

I need to sleep over it, but I guess in the grand scheme of things the number of multilingual Gosu projects is pretty small anyway.


mblumtritt:
> I'm not yet 100% sure if the new `Gosu.languages` method is really the best approach. 

I would like to see the method in the implemented form. Helping me to find out which languages are preferred is a very good idea!
Selecting the "correct language" if my app supports different languages should be part of the app logic and not the library. I would not like to limit the implementation to the scenario you described above.

Remark - because naming is important - I would choose
`Gosu.available_languages` to get the language set
`Gosu.preferred_language` to get the user preferred language (keep `Gosu.language` for backward compatibility as an alias)

jlnr:
Thanks for the feedback! Renaming the method sounds good. :+1:

A minor technical advantage of the current approach is that we already have an established pattern for handling array-of-string return values in FFI, but not for array-of-string parameters.

> Selecting the "correct language" if my app supports different languages should be part of the app logic and not the library.

Definitely sounds more in-character for Gosu to leave that to the game. And for some cultural reason, games (unlike any other app category) usually let the user override the language setting anyway.

Now if only I knew why the SDL 2.0.14 libraries for Windows somehow don't have the function I want to wrap...

jlnr:
@mblumtritt @cyberarm How about `user_languages` for the SDL_GetPreferredLocales wrapper? I wanted to emphasize that this is about what the user wants, something that `Gosu.language` never conveyed. Maybe I'm overthinking it, though.

PR #586, opened 2021-03-07
Add FFI error reporting

jlnr:
Rationale:
* For parity with SWIG Ruby/Gosu, ffi-gosu and gosu.cr need to throw helpful error messages when something goes wrong.
* The global Gosu_last_error() function should be easier to wrap than additional out parameters.
* For consistency, all Gosu_ functions (re)set the error, even ones like Gosu_milliseconds() which will hopefully never throw an exception.
  * ...except Color functions. Too weird?
* There is only an error message right now. Adding a Gosu_last_error_type() function later should be easy, but Gosu itself does not really throw many interesting exception classes, mostly runtime_error and invalid_argument.

Downsides:
* Now even harmless functions like Gosu_milliseconds should technically be wrapped in Ruby for error processing :( ... but otherwise, this is going to be a mess to document.

### DIFF
--- a/ffi/Gosu.cpp
+++ b/ffi/Gosu.cpp
@@ -1,31 +1,53 @@
 #include "Gosu_FFI_internal.h"
 
+std::string& Gosu_internal_error()
+{
+    static thread_local std::string error;
+    return error;
+}
+
+GOSU_FFI_API const char* Gosu_last_error(void)
+{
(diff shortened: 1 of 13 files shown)


cyberarm:
A quick test of `Gosu::Image.new("non-existant-font.png")` with
```ruby
if (err = Gosu.last_error)
  raise err
end
```
added to ffi-gosu Image constructor successfully raised a Rubyland error

PR #587, opened 2021-03-07
User languages patches

cyberarm:
* Drop `_get_` from `Gosu_get_user_languages` FFI function (@jlnr Is `_get_` intended?)
* Fix possible crash if `Gosu.user_languages` returns nil by using Ruby's safe navigation operator

### DIFF
--- a/ffi/Gosu.cpp
+++ b/ffi/Gosu.cpp
@@ -181,7 +181,7 @@ GOSU_FFI_API int Gosu_fps()
     return Gosu::fps();
 }
 
-GOSU_FFI_API void Gosu_get_user_languages(void function(void*, const char*), void* data)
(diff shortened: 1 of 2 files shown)


jlnr:
The `get` was intentional, but have no strong opinion about it. I can't stop seeing the FFI interface as a proper C library, and having a Gosu_user_languages method that (logically) returns an array felt weird.
If anything, the array-returning functions should probably be called `enumerate` instead of `get`.

Not sure about the safe navigation operator. In my opinion, the contract of `user_languages` should be to always return an array or throw an exception. Which reminds me that I completely forgot about documenting `user_languages` 🤦‍♂️

Can you please undo the safe operator? I'll merge the _get_ removal as it is more consistent with the rest of the FFI functions.

cyberarm:
Removed the safe navigation operator.

PR #588, opened 2021-03-13
Add support for getting display DPI

cyberarm:
Currently only tested on Linux atm.

### DIFF
--- a/ffi/Gosu.cpp
+++ b/ffi/Gosu.cpp
@@ -200,6 +200,13 @@ GOSU_FFI_API int Gosu_screen_height(Gosu_Window* window)
     });
 }
 
+GOSU_FFI_API double Gosu_dpi(Gosu_Window* window)
+{
+    return Gosu_translate_exceptions([=] {
+        return Gosu::dpi(window);
+    });
+}
+
 GOSU_FFI_API int Gosu_button_down(int id)
 {
     return Gosu_translate_exceptions([=] {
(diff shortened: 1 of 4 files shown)


jlnr:
The feature makes sense, but I wonder if percentage-based scaling factors aren't easier to understand.

I think under the hood Windows and X11 always express zoom in terms of DPI, but the user-facing Windows display settings call it 100%, 125%, 150% etc. macOS/iOS/GTK Linux are all about 100% vs 200% (less commonly also 300%).

Need to sleep over it :)

cyberarm:
@jlnr Have you had enough sleep on it? 😉

jlnr:
Yes and no. What I'm used to is Apple's API, specifically `[UIScreen mainScreen].scale` on iOS, which is typically 1.0, 2.0, or 3.0. Games can use this factor to load higher-resolution image assets on higher-resolution screens (as opposed to showing a larger viewport).

I think that use case makes sense, and I could live with both a floating-point scaling factor (because it makes the math easy) or an integer percentage value (which matches the Windows 10 UI).

At first I thought that SDL_GetDisplayDPI would just return "96" for a scaling factor of 100%. However, it seems that SDL_GetDisplayDPI is really reporting physical DPI, at least according to this SO answer: https://stackoverflow.com/a/38808654

Other than print design tools, should an app or game care about physical DPI, as opposed to how the user has configured their scaling factor? 🤔

There's an open SDL2 issue – I haven't read through it all, but I think what I want from SDL is not something that is supported right now: https://github.com/libsdl-org/SDL/issues/2119

I'm for putting this feature on hold, unless I'm missing the perfect use case for this.

jlnr:
Okay, um, it looks like SDL will finally have a consistent scaling model in SDL...3! https://github.com/libsdl-org/SDL/issues/7134#issuecomment-1555383062

Not sure if it even makes sense to touch this issue before that happens.

jlnr:
Specifically, SDL_GetDisplayDPI will be removed in SDL 3: https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md

Closing this PR for now. Hopefully Ubuntu 24.04 will already ship with SDL 3 so that we can start using it ASAP.

PR #591, opened 2021-03-22
Color refactoring

jlnr:


jlnr (commented):
@cyberarm I just wanted to fix warnings and ended up changing the C++ color class completely 😬 I think all the FFI stuff still works, but if you run into any issues, please let me know.

PR #594, opened 2021-04-24
Fix Gosu::Window#tick not raising errors from callbacks

cyberarm:
Reported on the forum: https://www.libgosu.org/cgi-bin/mwf/topic_show.pl?tid=1427


jlnr (approved):
LGTM, thanks!

jlnr:
Hmmm...but the failing test actually looks relevant. Not sure why it only fails with Ruby 3, though? 🤔

Issue #349, opened 2016-12-01
Video playback support


Dahrkael:
It would be nice to have a simple Video class supporting video playback, the same as Song for audio files.
We all know is not a must feature for low profile games, we all also know gosu is not used purely for games, but also for general multimedia apps.

It could use libtheoraplayer (https://github.com/AprilAndFriends/theoraplayer), which support Theora playback in all platforms and h264 in iOS (webm support incoming). Its abstracted so its very easy to set it up to use with OpenAL and OpenGL.

I have some old code for this which could be updated and used. It would need some OpenGL guru (im looking at you jlnr) to fix some Z ordering issues.

Dahrkael:
heres my old code for the old libtheoraplayer integration:
header: https://gist.github.com/Dahrkael/2ffad3ab0fb71a0faf222e6be4149c8c
source: https://gist.github.com/Dahrkael/6067e77e649f08c8e67d762ab840383a

As I see it, its pretty straightforward to do

jlnr:
Great library suggestion! Not sure how audio/video sync would work. But video playback is definitely a feature request that has come up a few times

jlnr:
@Quit on IRC has also managed to play YouTube in Gosu via Chrome Embedded Framework:

https://blorp.owndns.org/gefu2.png

Not sure if that isn't a more flexible solution? We get all of HTML! Weee! (But it seems troublesome in a lot of ways)

jlnr:
More CEF context: https://gist.github.com/Quit/d305fd1b6d4c9745ba4b9ffdbb3f8724

Quit:
I don't see the issue, personally. I have 146 MB of libraries, 919 kB of an executable, and around 50 kB of assets for the Gosu tutorial game. For me, that's a wholesome balance of everything.

Of course it's overkill. I've just wanted to see what it's like to integrate CEF into a game environment, I was interested in using it for Unity, but as that's kinda impossible, I didn't want all my research gone to waste - so I've thrown it at Gosu.

I really wouldn't recommend using this for anything, sans _maybe_ GUI bits - but even then, for Gosu's scope, that's probably overkill, too.

jlnr:
Mh, I don't really mind writing GUI code in Ruby, so I think CEF would be overkill for that.

But *if* CEF was really braindead easy to add to Gosu (not sure, haven't looked at the code/results yet :P), it could be used as a flexible hack to add things that only very few games need:

* video in a variety of formats (I think?)
* complex text formatting
[...]

PR #595, opened 2021-05-09
Add Gosu::Video

cyberarm:
Currently a stub.

An ffmpeg/libvlc backed implementation that'll emulate the `Gosu::Image` and `Gosu::Song` public interfaces.

`Gosu::Window::tick()` will likely need to call `Gosu::Video::update()` on all videos; A static vector in `Gosu::Video` can host and maintain the list.

The implementation is wrapped in an `#ifdef` to allow `Gosu::Video` to be an optional feature.

### DIFF
--- a/include/Gosu/Video.hpp
+++ b/include/Gosu/Video.hpp
@@ -1,15 +1,31 @@
-// TODO: Change to #ifdef when done implementing!
-#ifndef USE_FFMPEG
+#ifdef ENABLE_FFMPEG
 
 #pragma once
 
 #include <Gosu/Fwd.hpp>
+#include <Gosu/Graphics.hpp>
+#include <Gosu/GraphicsBase.hpp>
 #include <Gosu/IO.hpp>
+#include <Gosu/Math.hpp>
 #include <Gosu/Platform.hpp>
+#include <Gosu/Utility.hpp>
+
+#include <algorithm>
+#include <cassert>
+#include <cstdlib>
 #include <memory>
 #include <string>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #include <libavcodec/avcodec.h>
+#include <libavformat/avformat.h>
+
+#ifdef __cplusplus
+}
+#endif
 
 namespace Gosu
 {
--- a/include/Gosu/Video.hpp
+++ b/include/Gosu/Video.hpp
@@ -1,0 +1,61 @@
+// TODO: Change to #ifdef when done implementing!
+#ifndef USE_FFMPEG
+
+#pragma once
+
+#include <Gosu/Fwd.hpp>
+#include <Gosu/IO.hpp>
+#include <Gosu/Platform.hpp>
+#include <memory>
+#include <string>
+
+#include <libavcodec/avcodec.h>
+
+namespace Gosu
+{
+    class Video
+    {
+        struct Impl;
+        std::shared_ptr<Impl> pimpl;
+
+    public:
+        Video(const std::string& filename, unsigned image_flags);
+        Video(const std::string& filename, double width, double height, unsigned image_flags);
+
+        ~Video();
+
+        unsigned width() const;
+
+        unsigned height() const;
+
+        void draw(double x, double y, ZPos z, double scale_x, double scale_y, Color c,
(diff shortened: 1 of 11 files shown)


cyberarm:
@jlnr What do you think of this approach?

TODO:
- [ ] Prevent concurrent read/write of Image data
- [ ] Check for errors when creating `libvlc_` objects
- [ ] See why libvlc doesn't seem to like `:borderless` and if it can be fixed
- [ ] Add libvlc to /dependencies
- [ ] Link against libvlc on Windows and Mac

Video support is behind the `-DGOSU_ENABLE_VIDEO` flag

Working example:
```ruby
require_relative "ffi-gosu/lib/gosu"
require_relative "ffi-gosu/lib/gosu/video"


class Window < Gosu::Window
  def initialize
    super(800, 600, false, 16.6667, true, false)

    @video = Gosu::Video.new("/home/cyberarm/Downloads/big-buck-bunny_trailer.webm")

    pp @video.width, @video.height, @video.length, @video.position

    calculate_scale
  end

  def draw
    Gosu.draw_rect(0, 0, width, height, Gosu::Color::WHITE)
    @video.draw(0, 0, 0, @scale, @scale)
  end

  def update
    calculate_scale
  end

  def button_down(id)
    @video.play if id == Gosu::KB_SPACE
    @video.stop if id == Gosu::KB_LEFT_CONTROL
  end

  def calculate_scale
    @scale = [
      [
        width / @video.width.to_f,
        height / @video.height.to_f
      ].min,
      [
        width / @video.width.to_f,
        height / @video.height.to_f
      ].max
    ].min
  end
end

Window.new.show
```

Screenshot from 2021-05-10 15-43-00

include/Gosu/Video.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <vlc/vlc.h>
+
+#ifdef __cplusplus
+}
+#endif
```
jlnr, line 32: This does not have to be in the header, please move to .cpp

include/Gosu/Video.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+        double length() const;
+
+        double position() const;
+
+        void set_position(double position);
+
+        void update();
+
+        Gosu::Image* image();
```
jlnr, line 82: This member warrants a comment

jlnr (commented):
Wow, thanks for this. I didn't review all of it, but I left a couple of code comments.

My biggest worry with video support is shippability. ImageMagick seemed like a similar swiss army knife as VLC, but then packaging a Ruby/Gosu game with ImageMagick for Windows turned out to be a huge timesink. (I vaguely remember wrestling with ImageMagick "all in one" builds...)
Is VLC a single DLL that we can ship?

There was also an issue with example code for libtheora, I'm not sure how it compares to libvlc in terms of self-containedness: https://github.com/gosu/gosu/issues/349

(As long as it is hidden behind a compiler flag, this is not really a blocker. If video support is only a couple of extra files, why not!)

cyberarm:
> My biggest worry with video support is shippability. ImageMagick seemed like a similar swiss army knife as VLC, but then packaging a Ruby/Gosu game with ImageMagick for Windows turned out to be a huge timesink. (I vaguely remember wrestling with ImageMagick "all in one" builds...)
> Is VLC a single DLL that we can ship?

For libvlc we'd need to include `libvlc.dll`, `'libvlccore.dll`, and any required vlc plugins and codecs to playback video; We should probably only **_include_** playback support for ogg, webm, and av1, i.e. royality-free codecs, since they can be a landmine for commercial projects.
libvlc needs the `VLC_PLUGIN_PATH` environment variable set to locate plugins. It should be easy enough for users to add additional plugins/codecs that they may need.

> There was also an issue with example code for libtheora, I'm not sure how it compares to libvlc in terms of self-containedness: #349

The gists @Dahrkael linked to for libtheoraplayer are returning 404s and the waybackmachine doesn't have them.

jlnr:
Okay, that sounds good 👍 I was worried that we might need to ship a whole zoo of Unix-ish libraries.

PR #597, opened 2021-07-30
Clarify angle units for draw_rot

improbable-valentyn:
While it's pretty easy to find out experimentally whether it's degrees or radians, it is nicer to know in advance :)


jlnr:
Makes sense, thanks!

jlnr:
I think the intention was to have some central concepts explained here - https://github.com/gosu/gosu/wiki/Basic-Concepts

But that approach never really worked very well...so it's probably better to have a lot of redundancy in the RDoc.

Issue #599, opened 2021-10-01
Support for focus management


Rasunadon:
Add support for games to react to loss and gain of focus or possibly check if they are still focused. 
Idea is the ability to redraw the game (and not wait for next button press) on gaining of focus and gray out the drawn area when focus is lost. 

Something like
- Window#focus? – EDIT: question mark forgotten at first
- Window#gain_focus 
- Window#lose_focus 

Considered in https://discord.com/channels/433243560226193418/433243560226193420/758803411638157321

jlnr:
There is already a lose_focus callback in C++ because I wanted my iPhone game to auto-pause when the user jumps into a phone call. I guess adding a gain_focus counterpart would only be fair. I am not sure about #focus though.

jlnr:
Also, I think I might kill `needs_redraw?` in return. I just don't think it works very well. It might be easier to just `record{}` the whole game if there are few changes, and drawing that in #draw.

Rasunadon:
I added `focus` as it was mentioned in that old comment of yours... but I forgot to add the question mark at the end here – `focus?`. Yes, adding `gain_focus` event would be enough for my specific use case. Nox and Nic from forum seemed they could use that `focus?` checker (or `lose_focus` and `gain_focus` event handlers). I will check that `record{}`...

jlnr:
Forgot to close this, the callbacks have been released as part of Gosu 1.3.0. Hope this helps :+1:

PR #600, opened 2021-10-05
Added gain_focus callback, implemented lose_focus callback

cyberarm:
Addresses #599


cyberarm:
@jlnr I believe this patch is just about ready to be merged.
Only need to run `rake swig` and add another rename patch for `gain_focus` -> `protected_gain_focus`

PR #606, opened 2022-01-06
Updated SDL_sound, disabled and excluded midi support

cyberarm:


jlnr:
Thanks. Unfortunately, it seems SDL_sound is not compatible with the version of SDL2 that ships with Ubuntu 18.04:

```
/home/runner/work/gosu/gosu/dependencies/SDL_sound/SDL_sound_vorbis.c:64:13: warning: implicit declaration of function ‘SDL_exp’; did you mean ‘SDL_Log’? [-Wimplicit-function-declaration]
19
 #define exp SDL_exp
```

cyberarm:
@jlnr I removed the offending `#if !(defined(HAVE_LIBC) && defined(__WATCOMC__)) /* Watcom has issues... */` block which called the `#def exp SDL_exp`.

Looks like I also need to update the Visual Studio project... Good thing I got Windows booted on my main machine.

jlnr:
Thanks! How much manual tweaking was involved? Would it be possible to automate the updating process?

Although I hope that libsdl2-sound-dev (or so) will make its way into Ubuntu 22.04 so that we can just depend on it at some point.

cyberarm:
I only needed to:
1. fetch the repo
2. nuke the contents of the /dependencies/SDL_sound directory
3. move the contents of repo's /src into the SDL_sound folder
4. remove the midi file and library folder
5. edit SDL_sound_internal.h to change `#define SOUND_SUPPORTS_MIDI  1` to `#define SOUND_SUPPORTS_MIDI  0` (looks like this step could be skipped by setting this via cmake)
6. Remove the offending #ifdef block for `__WATCOMC__` that uses SDL 2.0.9 `SDL_exp` (this could also probably be simplified by setting `HAVE_LIBC` to 1 or `__WATCOMC__` to 0 via cmake)

PR #608, opened 2022-01-15
Replace OpenAL with mojoAL

cyberarm:


cyberarm:
@jlnr Do you want podspec updated to use mojoAL or leave that be for now?

jlnr:
Yes please, let's replace it all :)

cyberarm:
@jlnr I got all the CI passing. iOS and OSX should probably get manually checked to ensure they work as expected.
I tested sound on Windows and Linux, however I am unable to test iOS and OSX.

cyberarm:
iOS still uses OpenAL since mojoAL relies on SDL and Gosu's iOS implementation doesn't use SDL as far as I know.

jlnr (approved):
Had to add the header_mappings_dir so that installation through CocoaPods also works on macOS. Otherwise it seems to work just fine. Thanks so much!! Can't wait to never see oal-soft warnings in my Linux  terminals :)

PR #612, opened 2022-01-20
Fix Ruby 3.1 not adding DLL's to load path on Windows

cyberarm:
Ruby Installer's Ruby 3.1 changed the `RUBY_PLATFORM` string
See: https://rubyinstaller.org/2021/12/31/rubyinstaller-3.1.0-1-released.html


jlnr:
Thanks 🎆

PR #614, opened 2022-01-21
Statically link SDL2 (fixes #607)

### LINKED ISSUE #607: Gosu does not statically link SDL2 anymore on macOS?
I have probably subtly broken something, because `gosu.bundle` now references `/opt/homebrew/opt/sdl2/lib/libSDL2-2.0.0.dylib`, which it shouldn't. That means the bundle cannot be used with the portable Ruby.app anymore.

jlnr:
Need testing on an Intel Mac


jlnr:
Fixes #607 and #609

jlnr:
Removed the commits dropping support for Ruby 2.5 and 2.6 after realizing that current macOS still ships with 2.6.

PR #616, opened 2022-01-22
Added missing gain/lose_focus definitions to FFI window.h

cyberarm:


cyberarm:
I really should test ffi changes with crystal/mruby; cruby ffi is to forgiving...

jlnr (approved):

Issue #618, opened 2022-01-23
Gosu::Song using mojoAL is broken


cyberarm:
1. Gosu::Song will start stuttering after successfully playing a few songs all the way through.
2. Additionally starting a song and then stopping it after a few seconds to start playing it again sounds like it is repeating the previous fed samples.

The first issue seemingly is caused an `AL_INVALID_OPERATION` for `alBufferData` in https://github.com/gosu/gosu/blob/1658882080f8ba12d251b4a91688c3fcd0031c95/src/Audio.cpp#L91-L100

Which is triggered by mojoAL reference counter: https://github.com/gosu/gosu/blob/1658882080f8ba12d251b4a91688c3fcd0031c95/dependencies/mojoAL/mojoal.c#L4450-L4458

The second issue may also be caused by the same issue, I haven't dug around for it yet, only noticed it when implementing a quick test for the first issue so I didn't need to wait 5 - 10 minutes for it to trigger.

Quick trigger project: https://github.com/cyberarm/mojoal_stutter_trigger

cyberarm:
The second bug also appears to have the same cause: `alError stream_to_buffer-alBufferData : AL_INVALID_OPERATION`

cyberarm:
Here is what it sounds like

https://user-images.githubusercontent.com/581879/150707337-e73dfb45-79e4-4427-9295-1d7fcaa45bb7.mp4

cyberarm:
`AL_BUFFERS_QUEUED` (`alGetSourcei(source, AL_BUFFERS_QUEUED, &queued);`) may be returning an incorrect value
https://github.com/gosu/gosu/blob/1658882080f8ba12d251b4a91688c3fcd0031c95/dependencies/mojoAL/mojoal.c#L3724-L3726

https://github.com/gosu/gosu/blob/1658882080f8ba12d251b4a91688c3fcd0031c95/src/Audio.cpp#L145-L162

cyberarm:
Adding a call to `alSourcei(source, AL_BUFFER, 0);` inside of `stop()` fixes these issues, except for `AL_BUFFERS_QUEUED` possibly not returning the correct number of buffers to unqueue; Causing the first sample of `song.play` after `song.stop` to be from the previous run for that song.

Maybe that this will get resolved by: https://github.com/icculus/mojoAL/issues/6

```cpp
    void stop()
    {
        ALuint source = al_source_for_songs();

        alSourceStop(source);

        /* ADDED */
[...]

PR #619, opened 2022-01-24
Add call to alSourcei to Gosu::Song::Impl::stop() (ref: #618)

cyberarm:
This fixes the major issues of #618 with the caveat that mojoAL's `AL_BUFFERS_QUEUED` may be returning an incorrect value, causing a song that has been stopped and had `play()` called will have garbage data from that previous run for the first-ish sample.

https://github.com/icculus/mojoAL/blob/42035035b4e5debb6fe30e475683da2c30e51326/mojoal.c#L3976-L3977

### DIFF
--- a/src/Audio.cpp
+++ b/src/Audio.cpp
@@ -148,6 +148,8 @@ public:
 
         alSourceStop(source);
 
+        alSourcei(source, AL_BUFFER, AL_NONE);
+
         // Dequeue all buffers for this source.
         // The number of QUEUED buffers apparently includes the number of PROCESSED ones,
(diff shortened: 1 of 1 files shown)


jlnr:
Thanks. I haven't had time to dig into the mojoAL source, but this is definitely better than before. :+1:

icculus:
I believe this is fixed as of https://github.com/icculus/mojoAL/issues/6, but your workaround is valid and safe, too.

But if you happen to revisit this and find it's still broken in MojoAL without the workaround, please do let me know!

jlnr:
Thanks for the fix, and even more so for mojoAL! 🙏

Issue #479, opened 2018-11-23
Add retro option to Font


cyberarm:
![example](https://user-images.githubusercontent.com/581879/48959408-bfc7b000-ef2a-11e8-97b6-a3e09cf176a2.png)
Left: Non-retro
Right: Desired-ish result with retro

jlnr:
I have vague plans to solve this in a more generic way, namely, by passing a block to Font.new that is responsible for rendering characters into a Gosu::Image. That way, you could override some characters by loading them from files (e.g. Emoji), you could use `retro: true` to render them (I'm not sure if that's already possible with Image.from_text), etc.

fguillen:
I am also missing this :+1

jlnr:
Fixed in #624

PR #624, opened 2022-03-14
Add retro option to Font

cyberarm:
Addresses #479


jlnr:
I've pushed SWIG support for this feature, but somehow it doesn't seem to work? I'm probably missing something.

```ruby
require "gosu"

class W < Gosu::Window
  def draw
    @f ||= Gosu::Font.new(20, retro: true)
    @f.draw_text("rofl", 0, 0, 0, 60, 60)
  end
end

W.new(600, 600).show
```

ffi/Gosu_Font.cpp
```diff
@@ -38,6 +38,13 @@ GOSU_FFI_API unsigned Gosu_Font_flags(Gosu_Font* font)
     });
 }
 
+GOSU_FFI_API unsigned Gosu_Font_image_flags(Gosu_Font* font)
+{
+    return Gosu_translate_exceptions([=] {
+        return font->font.image_flags();
+    });
+}
```
jlnr: I'm removing this because MRI Ruby/Gosu doesn't have it either, even if it is somewhat inconsistent with C++ Gosu.

jlnr:
Looks like it was a SWIG issue, works now. Thanks!

Bildschirmfoto 2022-03-27 um 14 51 48

PR #625, opened 2022-03-27
Update SDL_sound, mojoAL, utf8proc, STB libraries

### LINKED ISSUE #592: Music stops too early
My cousin and I when we were making some games using Gosu noticed that the music (when using Gosu::Song) stops approximatively ~1 second before the end of the song and it restart. It's a bit annoying, since seamless musics are cut of 1sec before the end and loop and as so it's not seamless anymore.

I don't know what is the origin that issue sadly, I just know it happens with Gosu::Song.

### LINKED ISSUE #623: Gosu::Sample panning and speed broken under mojoAL
I only noticed this when trying to change a samples playback speed for my game jam project... 😞

jlnr:
Fixes #623
Fixes #592


src/BitmapIO.cpp
```diff
@@ -39,6 +39,26 @@ static bool is_bmp(Gosu::Reader reader)
     return magic_bytes[0] == 'B' && magic_bytes[1] == 'M';
 }
 
+/// Returns a copy of the given Gosu::Image with the alpha channel removed.
+/// Every pixel with alpha==0 will be replaced by Gosu::Color::FUCHSIA.
+static std::vector<Gosu::Color::Channel> bitmap_to_rgb(const Gosu::Bitmap& bmp)
```
jlnr, line 44: Required to work around the fact that stb_image_write will now actually write BMP files with an alpha channel: https://github.com/nothings/stb/commit/b14a80784c689979b75f6e8929a7d7b4b4a86678

...and for compatibility reasons, Gosu should keep writing 24-bit BMP files

PR #630, opened 2022-08-12
Add Clipboard Support

cyberarm:
Needs to be tested and add swig support

### DIFF
--- a/include/Gosu/Input.hpp
+++ b/include/Gosu/Input.hpp
@@ -109,6 +109,6 @@ namespace Gosu
         void set_text_input(TextInput* input);
 
         static std::string clipboard();
-        static void set_clipboard(std::string text);
+        static void set_clipboard(const std::string& text);
     };
 }
--- a/include/Gosu/Input.hpp
+++ b/include/Gosu/Input.hpp
@@ -108,7 +108,7 @@ namespace Gosu
         //! Sets the currently active TextInput, or resets it to the nullptr.
         void set_text_input(TextInput* input);
 
-        static void set_clipboard(std::string text);
         static std::string clipboard();
+        static void set_clipboard(std::string text);
     };
 }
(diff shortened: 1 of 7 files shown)


cyberarm:
⚠️Forgot: Free string with `SDL_free`
 https://wiki.libsdl.org/SDL_GetClipboardText

include/Gosu/Input.hpp
```diff
@@ -107,5 +107,8 @@ namespace Gosu
         TextInput* text_input() const;
         //! Sets the currently active TextInput, or resets it to the nullptr.
         void set_text_input(TextInput* input);
+
+        static void set_clipboard(std::string text);
+        static std::string clipboard();
```
jlnr, line 112: Usually the setter comes after the getter...although that's probably not a super consistent convention.

src/Input.cpp
```diff
@@ -678,4 +678,14 @@ void Gosu::Input::set_text_input(TextInput* text_input)
...
 }
 
+void Gosu::Input::set_clipboard(std::string text)
+{
+    SDL_SetClipboardText(text.c_str());
+}
+
+std::string Gosu::Input::clipboard()
+{
+    return std::string(SDL_GetClipboardText());
```
jlnr, line 688:
```suggestion
    std::shared_ptr<char> clipboard{SDL_GetClipboardText(), SDL_free};
    return std::string{clipboard.get()};
```

jlnr (commented):
Thanks ✌️

cyberarm:
@jlnr This should be ready, except for the SWIG bits.

jlnr (approved):

jlnr:
If you want, you can add a unit test for it. (I didn't even get to test it manually...) Otherwise, LGTM, and thanks!

cyberarm:
@jlnr I added a test for `Gosu.clipboard`, if CI passes it should be good to merge! 🥳

jlnr:
Thanks again, merged. :+1:

PR #637, opened 2023-01-10
Expose SDL window pointer

cyberarm:
Added `Gosu::Window::sdl_window() `and `Gosu_Window_sdl_window(Gosu_Window*)` to expose `Gosu::shared_window()` SDL window pointer to applications


jlnr:
This doesn't need SWIG wrapping because I'm not sure what MRI users would do with this handle, but I might need to %ignore it to prevent it from appearing when `rake swig` is run.

cyberarm:
@jlnr Have you had time to think more about this?

jlnr (approved):
Merged master into this branch. Sorry if you prefer rebase, I have too much muscle memory from various projects now :) I have opted to %ignore the new field in the SWIG version because it was indeed exposed as an opaque _p_void wrapper that I don't think would have helped anyone. Let's wait for the MRI version of this until we have experience from mruby?

jlnr:
Please merge this if you still have a need for this. If not, then as usual I'd prefer to defer all features until a game (or UI project :)) uses (and passively tests) it.

PR #639, opened 2023-03-26
Use SDL_GameControllerSetPlayerIndex and SDL_JoystickSetPlayerIndex 

cyberarm:
Use `SDL_GameControllerSetPlayerIndex` and `SDL_JoystickSetPlayerIndex` to have SDL set gamepad LED(s), refactored a bit to use the `event->jdevice.which` instead of iterating over known joysticks: fixes `wireless xbox 360 controller` from appearing twice if connected after program starts.

> This function is available since SDL 2.0.12.

### DIFF
--- a/src/Input.cpp
+++ b/src/Input.cpp
@@ -140,39 +140,38 @@ struct Gosu::Input::Impl : private Gosu::Noncopyable
                 }
                 int gamepad_slot = -1;
                 int joystick_instance_id = -1;
+                int i = e->jdevice.which; // SDL2 device_index
 
-                // Loop through attached gamepads as e->jdevice.which cannot be trusted (always 0)
-                for (int i = 0; i < SDL_NumJoysticks(); i++) {
-                    // Prefer the SDL_GameController API...
-                    if (SDL_IsGameController(i)) {
-                        if (SDL_GameController *game_controller = SDL_GameControllerOpen(i)) {
-                            gamepad_slot = available_gamepad_slot_index();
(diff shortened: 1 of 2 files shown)


cyberarm:
@jlnr Do we still need to support Ubuntu 18.04? 20.04 and 22.04 LTS's are out and 18.04 is End of Life in May.

> https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices

jlnr:
Only the last two LTS versions -> 20.04 and 22.04 👍

cyberarm:
Alright, looks like we get SDL v2.0.10 on 20.04... Time for some `#if`

cyberarm:
@jlnr There we are, builds passing now.

src/Input.cpp
```diff
@@ -140,39 +140,42 @@ struct Gosu::Input::Impl : private Gosu::Noncopyable
                 }
                 int gamepad_slot = -1;
                 int joystick_instance_id = -1;
+                int i = e->jdevice.which; // SDL2 device_index
 
-                // Loop through attached gamepads as e->jdevice.which cannot be trusted (always 0)
```
jlnr, line 144: This comment was introduced here: https://github.com/gosu/gosu/pull/543

Soo...does this mean we can actually trust e->jdevice.which now? Was it only broken in the 18.04 version?

jlnr (approved):
I wish I had the time to thoroughly test this 😬 But using the new function sounds like a great idea.

PR #648, opened 2023-07-15
More CMake work (goal: make CLion's coverage reporting work)

jlnr:
Let's see if making Gosu non-SHARED breaks anything. The thing I don't quite understand is how the generated GosuTargets.cmake is supposed to find its dependencies.


jlnr:
At least this now works, yay.

Capture d’écran 2023-07-15 à 18 24 02

PR #662, opened 2023-08-02
Refactor Window.cpp, especially threading with regards to OpenGL

jlnr:
Still needs more tests.


src/Texture.cpp
```diff
@@ -48,8 +48,7 @@ Gosu::Texture::Texture(int width, int height, bool retro)
 
 Gosu::Texture::~Texture()
 {
-    ensure_current_context();
-
+    const auto lock = lock_gl_context();
     glDeleteTextures(1, &m_tex_name);
```
jlnr, line 52: Should probably swallow exceptions here.

test/DrawableTests.cpp
```diff
@@ -116,26 +116,32 @@ TEST_F(DrawableTests, multithreaded_stress_test)
...
                 }
+            } catch (const std::exception& e) {
+                GTEST_FAIL() << e.what();
             }
         });
     }
 
     for (auto& thread : threads) {
         thread.join();
     }
```
jlnr, line 146: Would be good to test that textures allocated from background threads can be read from the main thread (i.e. sharing works).

PR #664, opened 2023-08-07
Replace SWIG wrapper with github.com/cyberarm/ffi-gosu

jlnr:
I am a little shocked that my very first attempt at "rake gem; gem install ...; gosu-examples" already worked 🤔

Now to fix all the backward compatibility shims...


ext/gosu/extconf.rb
```diff
@@ -101,9 +98,6 @@
   end
 
   pkg_config "sdl2"
-  pkg_config "vorbisfile"
-  pkg_config "sndfile"
-  pkg_config "libmpg123"
```
jlnr, line 106: Not sure why they were still in there, but now I'm worried I might have removed too many unused audio libraries form the "Getting Started on Linux" page 🤔

lib/gosu/window.rb
```diff
@@ -0,0 +1,312 @@
...
+    attach_function :Gosu_Window_set_button_down,          [:pointer, :_callback_with_unsigned, :pointer],             :void
+    attach_function :Gosu_Window_set_button_up,            [:pointer, :_callback_with_unsigned, :pointer],             :void
+    attach_function :Gosu_Window_set_gamepad_connected,    [:pointer, :_callback_with_unsigned, :pointer],             :void
+    attach_function :Gosu_Window_set_gamepad_disconnected, [:pointer, :_callback_with_unsigned, :pointer],             :void
+    attach_function :Gosu_Window_set_drop,                 [:pointer, :_callback_with_string, :pointer],               :void
+    attach_function :Gosu_Window_set_needs_redraw,         [:pointer, :_callback_returns_bool, :pointer],              :void
+    attach_function :Gosu_Window_set_needs_cursor,         [:pointer, :_callback_returns_bool, :pointer],              :void
+    #attach_function :Gosu_Window_set_capture_cursor,       [:pointer, :_callback_returns_bool, :pointer],              :void
+    #attach_function :Gosu_Window_set_hit_test,             [:pointer, :_callback_hit_test_returns_unsigned, :pointer], :void
```
jlnr, line 25:
```suggestion
```

lib/gosu/version.rb
```diff
@@ -0,0 +1,3 @@
+module Gosu
+  GEM_VERSION = "0.0.1"
```
jlnr, line 2: Name should to be changed and updated by `rake set_version`

lib/gosu/color.rb
```diff
@@ -0,0 +1,175 @@
...
+    attach_function :Gosu_Color_set_saturation, [:uint32, :double], :uint32
+    attach_function :Gosu_Color_set_hue,        [:uint32, :double], :uint32
+
+    attach_function :Gosu_Color_bgr,  [:uint32], :uint32
+    attach_function :Gosu_Color_abgr, [:uint32], :uint32
+    attach_function :Gosu_Color_argb, [:uint32], :uint32
+
+    # Gosu::Color is stored as an unsigned int in C
+    # this is used for {Gosu::Image#to_blob} and {Gosu::Image.from_blob}
+    SIZEOF = 4
```
jlnr, line 35: Note to self: Why does this need to be exposed in Ruby?

lib/gosu/color.rb
```diff
@@ -0,0 +1,175 @@
...
+    def initialize(*args)
+      case args.size
+      when 1
+        @__color = Gosu_Color_create(args.first)
+        Gosu.check_last_error
+      when 4
+        @__color = Gosu_Color_create_argb(args[0].clamp(0, 255), args[1].clamp(0, 255), args[2].clamp(0, 255), args[3].clamp(0, 255))
+        Gosu.check_last_error
+      else
+        raise ArgumentError
```
jlnr, line 70: Should be more descriptive

lib/gosu.rb
```diff
@@ -17,10 +17,369 @@
...
+    Gosu_fps().tap { Gosu.check_last_error }
+  end
+
+  def self.flush
+    Gosu_flush()
+    Gosu.check_last_error
+  end
+
+  def self.milliseconds
+    Gosu_milliseconds().tap { Gosu.check_last_error }
```
mblumtritt, line 100: Using #tap is very time consuming because a Proc is created. You should avoid this…

lib/gosu.rb
```diff
@@ -17,10 +17,369 @@
...
+
+    raise "Block not given!" unless block
+
+    if z
+      Gosu_gl_z(z, block)
+      Gosu.check_last_error
+    else
+      Gosu_gl(block)
+      Gosu.check_last_error
+    end
```
mblumtritt, line 132:
```suggestion
    z ? Gosu_gl_z(z, block) :  Gosu_gl(block)
    Gosu.check_last_error
```

lib/gosu.rb
mblumtritt: Maybe `Gosu.check_last_error` accepts optional a parameter which will be returned when no error was found. This seems to help and clarify your code in most cases…

jlnr: Thanks. I didn't know about the Proc creation, but I also had a feeling that this is doing too much on the Ruby side.

jlnr: In this very specific method it's probably better to skip the error handling altogether, and document it as being noexcept.

lib/gosu/color.rb
```diff
@@ -0,0 +1,175 @@
...
+      gl == (other.is_a?(Numeric) ? other : other&.gl)
+    end
+
+    def <(other)
+      gl < (other.is_a?(Numeric) ? other : other&.gl)
+    end
+
+    def >(other)
+      gl > (other.is_a?(Numeric) ? other : other&.gl)
+    end
```
jlnr, line 156: Can these be delegated to <=> as in C++20?

lib/gosu/color.rb
```diff
@@ -0,0 +1,175 @@
...
+      when 4
+        @__color = Gosu_Color_create_argb(args[0].clamp(0, 255), args[1].clamp(0, 255), args[2].clamp(0, 255), args[3].clamp(0, 255))
+        Gosu.check_last_error
+      else
+        raise ArgumentError
+      end
+    end
+
+    def alpha
+      Gosu_Color_alpha(@__color).tap { Gosu.check_last_error }
```
jlnr, line 75: Probably better to skip error checks for Color methods as well.

lib/gosu/compat.rb
```diff
@@ -1,196 +1,48 @@
...
-      if args.size == 7
-        initialize_without_window args[1], tileable: args[2], rect: args[3..-1]
-      else
-        initialize_without_window args[1], tileable: args[2]
+  class Window
+    # Compat code taken from gosu/gosu
+
+    %w(update draw needs_redraw? needs_cursor? capture_cursor? hit_test gain_focus lose_focus
+      lose_focus button_down button_up gamepad_connected gamepad_disconnected drop close).each do |callback|
+      define_method "protected_#{callback}" do |*args|
```
jlnr, line 25: Gosu's SWIG wrapper used to call protected_foo from the C++ directory stuff. Whatever this is trying to do needs to be done in a different way now.

jlnr (commented):
First the good news: Rubystein 3D still seems to be exactly as fast as with the CRuby version. I've added `10.times` around its draw method and am still getting 20 FPS. It's almost creepy how seamless this change is? Am I doing something wrong?

lib/gosu/color.rb
mblumtritt: The Ruby module Comparable does the trick!

jlnr: Thanks, done.

lib/gosu.rb
jlnr: I still went with `if else end` because I prefer to use ?: only for expressions, but I have DRY'd the check_last_error call, thanks.

lib/gosu.rb
jlnr: I've changed check_last_error as you suggested. I left the error checking even in methods that cannot possibly throw because I was afraid it would be confusing to document which methods are noexcept, and which ones aren't.

However, I have removed the error checks from all `release` methods because C++ destructors really must never throw, and that's a rule that should be easy to follow in the wrapper code.

lib/gosu/compat.rb
jlnr: Ah, window.rb still uses a similar mechanism even with FFI. Need to look deeper into this.

lib/gosu/color.rb
jlnr: Hmmm...leaving it for now, otherwise it just gets too confusing.

lib/gosu.rb
```diff
@@ -17,10 +17,374 @@
...
+  # It might be useful to also allow loading Gosu installed as a system library, e.g.:
+  # "/usr/local/lib/libgosu-ffi.dylib", # Homebrew on macOS (Intel) or manual installation
+  # "/opt/homebrew/lib/libgosu-ffi.dylib", # Homebrew on macOS (Apple Silicon)
+  # ...or built locally, for testing without having to re-install the gem:
+  # File.expand_path("../../gosu/build/ffi/libgosu-ffi.so", __dir__),
+  # File.expand_path("../../gosu/lib64/gosu-ffi.dll", __dir__),
+  library = "gosu.#{RbConfig::MAKEFILE_CONFIG['DLEXT']}"
+  candidates = ["#{__FILE__}/..", "#{__FILE__}/../../ext/gosu"]
+  candidates.prepend(Gem.loaded_specs['gosu'].extension_dir) if Gem.loaded_specs['gosu']
+  ffi_lib(candidates.map { |dir| File.expand_path(library, dir) })
```
jlnr, line 38: Inspired by https://github.com/dbalatero/levenshtein-ffi/pull/11

But I don't want to adjust the ffi_lib paths in every single RB file here. I guess there should only be one central ffi_lib statement.

lib/gosu/constants.rb
```diff
@@ -0,0 +1,332 @@
...
+    "IF_TILEABLE_TOP",
+    "IF_TILEABLE_RIGHT",
+    "IF_TILEABLE_BOTTOM",
+    "IF_TILEABLE",
+    "IF_RETRO",
+  ]
+
+  constants.each do |const|
+    attach_variable :"#{const}", :"Gosu_#{const}", :uint32
+    Gosu.const_set(const, Gosu.send(:"#{const}"))
```
mblumtritt, line 320: You don't need conversions if you already fill in your `constants` Array with Symbols instead of Strings:

```suggestion
  constants.each do |const|
    attach_variable const, :"Gosu_#{const}", :uint32
    Gosu.const_set(const, Gosu.send(const))
```

lib/gosu/constants.rb
```diff
@@ -0,0 +1,332 @@
...
+
+    "GP_3_LEFT_STICK_X_AXIS",
+    "GP_3_LEFT_STICK_Y_AXIS",
+    "GP_3_RIGHT_STICK_X_AXIS",
+    "GP_3_RIGHT_STICK_Y_AXIS",
+    "GP_3_LEFT_TRIGGER_AXIS",
+    "GP_3_RIGHT_TRIGGER_AXIS",
+  ]
+
+  constants = constants + [
```
mblumtritt, line 282: Is there any reason to split them in two arrays?
And why not just

```suggestion
  ] +
  [
```

or
```suggestion
  ]

  constants += [
```

lib/gosu/window.rb
```diff
@@ -0,0 +1,284 @@
...
+        button_up: [:id],
+        gamepad_connected: [:id],
+        gamepad_disconnected: [:id],
+        drop: [:filename],
+        needs_redraw?: [],
+        needs_cursor?: [],
+        close: [],
+        gain_focus: [],
+        lose_focus: [],
+      }.each do |callback, args|
```
mblumtritt, line 89: You define Symbol keys but convert them into Strings later on in any use case. Why not using strings in the first place?

jlnr: It looks nicer, I guess. I tried to refactor this and the counterpart in compat.rb (that defines the protected_ methods), but I think I'll move this to a later PR, as long as the tests pass.

lib/gosu/constants.rb
```diff
@@ -0,0 +1,326 @@
+module Gosu
+  # uint32
+  constants = [
+    "KB_ESCAPE",
```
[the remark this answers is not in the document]
jlnr: Not in this PR, though.

lib/gosu/text_input.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    def selection_start=(int)
+      Gosu_TextInput_set_selection_start(__pointer, int)
+      Gosu.check_last_error
+    end
+
+    def filter(text)
+      return text
+    end
```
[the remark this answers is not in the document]
jlnr: The idea is that C++ calls `window.current_text_input.filter()` when text is being entered, and this method can be overridden. This is the default implementation, which does nothing.

This mechanism was really awkward to get to work with FFI. Also something to revisit in a PR about protected_ callbacks.

jlnr:
One check was mysteriously failing on Ubuntu, but only with Ruby 2.6. Two other tests (one C++, one Ruby) are failing on my Ubuntu machine as well. I don't think that this necessarily has anything to do with the FFI switch, I suspect that something is just generally broken :(

PR #674, opened 2023-09-14
Fix and test subimage, loading images from rects, and BLOB roundtrips

jlnr:


jlnr (commented):
`TestImage#test_image_insert` is much faster now. I suspect that `blob.write_array_of_type` was the bottleneck.

jlnr:
FYI @cyberarm These are probably breaking changes for gosu.cr.

PR #681, opened 2024-08-02
Initial stab at Swift support

pusewicz:
Hey Gosu team!

I've been playing around a bit with Swift and Gosu, and was wondering if this is something that would be worth pursuing.

The basic structs can be imported no problem and can be used directly in Swift.

What currently does not "just work" is the `Window` class for example as it's using reference symantics and would need marking as such in Swift (or using a C bridge, like the one in FFI).

Please let me know your thoughts and feel free to close this if this is definitely a do-not-want. Thanks!

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.22.1)
 
+add_subdirectory(Swift)
 add_subdirectory(TextInput)
 add_subdirectory(Tutorial)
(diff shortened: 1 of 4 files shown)


jlnr:
Ooh, I was wondering how this is not a re-implementation of the ObjC wrapper as in https://github.com/gosu/motion-gosu, but you are using the new Swift/C++ bridge. Nice to see that it generally works.

> the Window class for example as it's using reference symantics and would need marking as such in Swift 

If it's just a matter of adding a few annotations in the header file, then of course I'd be willing to add them. Window and TextInput have always been the hardest classes to wrap :( I'm also curious how error handling works, does it rely on C++ noexcept annotations?

PR #685, opened 2026-01-09
SDL2 to SDL3

cyberarm:


jlnr:
Ooh, interesting. Don't we also need sdl2-compat so that our code keeps working? I will be traveling for the next month but March will be my first vacation in ages, let's see if I can finally merge some things here.

cyberarm:
>  Don't we also need sdl2-compat so that our code keeps working? 

Nope, SDL3 is largely the same as SDL2 (with some really nice built-ins and new features) and mainly needed some function calls renamed and a few arguments changed.
Got my W3D Hub Linux Launcher ruby app running under SDL3 on windows with only some weirdness with fullscreen toggling and rendering getting washed out. 🐛 

> let's see if I can finally merge some things here.

👍

I'm wanting to make it so that the rubygem ships with precompiled binaries on most platforms (needs CI/CD updated/set up) and the C(++) library to not depend on the system installed SDL version so we're not stuck on ubuntu's old lts versions.
May be to much to ask, but we'll see how things progress 🙂

jlnr:
Well that would be a full roundtrip to where Gosu started, with all libraries bundled at least on macOS and Windows. :D

Regarding Linux, maybe it would be better to invest in a Flatpak setup? I guess Flathub and Steam (as a Windows EXE) are the best targets for Gosu games as long as there is no WASM support. In any case, I wouldn't mind dropping Ubuntu LTS-1 support if it helps. It looks like SDL3 is available as a backport for 24.04 too.

jlnr:
Sorry for the conflicts in #688, I am trying to tackle the non-SDL3 bits in this PR to shrink it a bit...

jlnr:
Wow, I thought we'd have more time with this, but now Homebrew has removed the sdl2 formula. I mean, sdl2-compat exists... but that still invalidates all of Gosu's installation instructions etc. :(

jlnr-tw:
https://github.com/gosu/gosu/pull/692 will again have many conflicts with this PR. One question I've been pondering: Isn't it also time to drop 32-bit support for Windows? It looks like RubyInstaller has already done so. Although I guess there won't be any savings because arm64 is just around the corner to replace it.

jlnr:
Oops, I should have made a backup of your branch instead of force-pushing. If you spot anything that I've clobbered while resolving conflicts, please let me know.

Now trying to get this to green on the CI...will have access to a Windows machine again later today.

jlnr:
@cyberarm Done & merged. Thanks a lot for getting the SDL3 party started 🙏

PR #687, opened 2026-06-21
Fix GitHub CI

jlnr-tw:
The versions of Ruby and the iOS simulator were outdated. Not sure if Actions will run here? (Posting this from my "less secure" GitHub account)


jlnr:
Hmm, we had a segmentation fault with Ruby 4 on Windows but then it went away without any specific changes. Let's see if it will happen again...

PR #688, opened 2026-06-21
Update SDL to 2.32.0, fix corresponding rake task

jlnr-tw:
The actual changes are the ones in rake/update_dependencies.rb, heavily inspired by #685.


rake/update_dependencies.rb
```diff
@@ -25,23 +23,28 @@
   if response.status == 200
     array = JSON.parse(response.body, symbolize_names: true)
     selected_release = array.find do |release|
-      !release[:draft] && !release[:prerelease] && Integer(release[:name].split(".")[1]).even?
+      next false if release[:draft] or release[:prerelease]
+
+      next false unless release[:assets].any? { |asset| asset[:name].end_with?("-VC.zip") }
```
jlnr-tw: I am not a big fan of ! and && so I have used `next false` here which seems to do the trick.

